PR #6, opened 2021-04-20
Create CONTRIBUTING.md

Richard-Gist:


CONTRIBUTING.md
```diff
@@ -0,0 +1 @@
+
```
Tyler-Keith-Thompson, line 1: What is up with this line?

Richard-Gist: It's the starting line for an empty file.

PR #10, opened 2021-04-26
Orchestration engine

### LINKED ISSUE #9: Extreme type safety!
Workflow has some sort of type safety at the moment and that's very helpful. We can take it much, much farther with the notion that compile-time errors are preferred. 

For example every FlowRepresentable can declare both an `Input` and `Output`, then when you create a workflow you can enforce (with a fluent API) that the next thing being presented has the same kind of Input that the last thing outputs.

We could not have type safety and just make everyone deal with "Any" for the rest of their lives, but that seems cruel.

### Additional context
It's already done.....mostly.

Tyler-Keith-Thompson:
## Checklist:
- [x] Did you comply with our [styleguide](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/STYLEGUIDE.md)?
- [x] Is there [adequate test coverage](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#test-etiquette) for your new code?
- [x] Is the [static analysis tool](https://sonarcloud.io/dashboard?id=Tyler-Keith-Thompson_Workflow) reporting 0 issues?
- [x] Did you [update the documentation](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#documentation)?
- [x] Did you [update the sample app](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#sample-app)?
- [x] Do we need to [increment the minor/major version](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#versioning)?
- [x] Did you [change the public API](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#public-api)?
- [x] Have you done everything you can to make sure that errors that can occur are compile-time errors, and if they have to be runtime do you have adequate tests and documentation around those runtime errors? [For more details](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#errors).

----

### If Applicable:

----

- [x] Did you deprecate (rather than remove) any old methods/variables/etc? [Our philosophy for deprecation](https://github.com/Tyler-Keith-Thompson/Workflow/blob/master/CONTRIBUTING.md#deprecation).


Workflow/Sources/Workflow/Models/AnyWorkflow.swift
```diff
@@ -0,0 +1,187 @@
...
+            let nextMetadata = nextNode.value
+            let flowRepresentable = nextMetadata.flowRepresentableFactory()
+            flowRepresentable.workflow = self
+            flowRepresentable.proceedInWorkflowStorage = { passedArgs = $0 }
+
+            let shouldLoad = flowRepresentable.shouldLoad(with: passedArgs)
+
+            defer {
+                guard let instance = instances.first?.traverse(nextNode.position) else { fatalError("Internal state of workflow completely mangled somehow...") }
+                let persistance = nextMetadata.calculatePersistance(passedArgs)
```
Richard-Gist, line 51: Leaving a note here so we don't forget.  We need to fix the spelling of 
`Persistance` to `Persistence`
```suggestion
                let persistence = nextMetadata.calculatePersistence(passedArgs)
```
everywhere.  This is a big change that would break the API so we should do it in this version ramp instead of making 2.

Richard-Gist (changes requested):
Marking this as a requested change so we don't forget to address it.

Workflow/Sources/Workflow/Models/AnyWorkflow.swift
```diff
@@ -0,0 +1,187 @@
...
+
+        return firstLoadedInstance
+    }
+
+    /// abandon: Called when the workflow should be terminated, and the app should return to the point before the workflow was launched
+    /// - Parameter animated: A boolean indicating whether abandoning the workflow should be animated
+    /// - Parameter onFinish: A callback after the workflow has been abandoned.
+    /// - Returns: Void
+    /// - Note: In order for this to function the workflow must have a presenter, presenters must call back to the workflow to inform when the abandon process has finished for the onFinish callback to be called.
+    public func abandon(animated: Bool = true, onFinish:(() -> Void)? = nil) {
```
Richard-Gist, line 84: I'm not sure how to do this, but this abandon should not take in an animated parameter.

Richard-Gist (changes requested):
Marking this as requested change.  The other instance of using animated should have also been a requested change.

Richard-Gist (changes requested):

Workflow/Sources/Workflow/Models/AnyWorkflow.swift
```diff
@@ -0,0 +1,187 @@
...
+    public typealias ArrayLiteralElement = AnyFlowRepresentable.Type
+    internal var instances = LinkedList<AnyFlowRepresentable?>()
+    internal var orchestrationResponder: AnyOrchestrationResponder?
+
+    public func applyOrchestrationResponder(_ orchestrationResponder: AnyOrchestrationResponder) {
+        self.orchestrationResponder = orchestrationResponder
+    }
+
+    @discardableResult public func launch(withLaunchStyle launchStyle: LaunchStyle = .default,
+                                          onFinish: ((Any?) -> Void)? = nil) -> LinkedList<AnyFlowRepresentable?>.Element? {
```
Tyler-Keith-Thompson, line 22: Should all these `onFinish` blocks actually take in an `AnyWorkflow.PassedArgs` enum? Consumers can then make deliberate decisions if no data was passed forward at the end of a workflow. Right now consumers cannot distinguish between `nil` being a legitimately passed value and nothing being passed. That distinction may be very important

Workflow/Tests/WorkflowTests/OrchestrationResponder/PersistenceTests.swift
```diff
@@ -0,0 +1,239 @@
...
+//
+//  Created by Tyler Thompson on 11/25/20.
+//
+
+import Foundation
+import XCTest
+
+@testable import Workflow
+
+class PersistenceTests: XCTestCase {
```
Tyler-Keith-Thompson, line 13: Wild guess that these shouldn't be commented out. They should either be removed or explicitly dealt with by other tests

Workflow/Sources/Workflow/Models/AnyWorkflow.swift
```diff
@@ -0,0 +1,187 @@
...
+        launch(passedArgs: .none, withLaunchStyle: launchStyle, onFinish: onFinish)
+    }
+
+    @discardableResult public func launch(with args: Any?,
```
Tyler-Keith-Thompson, line 26: Is it possible to use the `Workflow` class and get a type-safe launch? The args would have to be the *first* FlowRepresentable.WorkflowInput value. It may be impossible or simply not reasonable, but it would be quite nice as a consumer I think.

Workflow/Sources/Workflow/Protocols/FlowRepresentable.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// instance: A method to return an instance of the `FlowRepresentable`
+    /// - Returns: `AnyFlowRepresentable`. Specifically a new instance from the static class passed to a `Workflow`
+    /// - Note: This needs to return a unique instance of your view. Whether programmatic or from the storyboard is irrelevant
+    static func instance() -> Self
+
+    /// shouldLoad: A method indicating whether it makes sense for this view to load in a workflow
+    /// - Parameter args: Note you can rename this in your implementation if 'args' doesn't make sense.
+    /// - Returns: Bool
+    /// - Note: This method is called *before* your view loads. Do not attempt to do any UI work in this method. This is however a good place to set up data on your view.
+    mutating func shouldLoad(with args: WorkflowInput) -> Bool
```
Tyler-Keith-Thompson, line 34: There is very likely some clever way to combine the concept of `instance` and `shouldLoad` into an optional initializer that takes in the WorkflowInput. A `shouldLoad` of `false` would be that initializer returning nil. 

This would make life fantastic in SwiftUI. It'd be important to make sure that it continued to work well with UIKit but UIKit actually has it's own optional initializer for much this purpose. Look at [this article](https://www.hackingwithswift.com/example-code/uikit/how-to-use-dependency-injection-with-storyboards) for more details.

WorkflowExample/WorkflowUIKitTests/UIKitConsumerTests.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  UIKitPresenterTests.swift
+//  UIKitConsumerTests.swift
```
Tyler-Keith-Thompson, line 2: This file is HUGE, it'd be a good idea to break the tests into different logical file groups

Tyler-Keith-Thompson:
Idea: We could have a `Passthrough` type that's a little different than `Never`. Whereas `Never` means "I don't take in a value" or "I don't pass out a value". `Passthrough` could mean "I'll forward on anything that gets passed to me automatically`. 

Hypothetical: Imagine a workflow wherein you've got a terms and conditions `FlowRepresentable`. Now it knows whether terms have been accepted or not, and it will make the decision to abandon whatever workflow it is in if the terms are not accepted. However if terms *are* accepted it might send that information to a server then move forward. 

If that termsAndConditions had a `Passthrough` type for its input and output it would be even easier to stick those terms and conditions in *any* workflow while maintaining type safety of that flow.

Workflow/Sources/Workflow/Models/AnyWorkflow.swift
Richard-Gist: Yes they should and good job doing that.

Workflow/Tests/WorkflowTests/OrchestrationResponder/PersistanceTests.swift
```diff
@@ -0,0 +1,239 @@
+//
+//  PersistanceTests.swift
```
Richard-Gist: We should uncomment these tests, or delete them.

Richard-Gist (changes requested):

Richard-Gist (dismissed):
Looks good

Richard-Gist (approved):

PR #15, opened 2021-05-03
Init example

Tyler-Keith-Thompson:
## Checklist:
- [x] Did you [update the sample app](https://github.com/wwt/Workflow/blob/main/CONTRIBUTING.md#sample-app)?
- [x] Do we need to [increment the minor/major version](https://github.com/wwt/Workflow/blob/main/CONTRIBUTING.md#versioning)?
- [x] Did you [change the public API](https://github.com/wwt/Workflow/blob/main/CONTRIBUTING.md#public-api)?
- [x] Have you done everything you can to make sure that errors that can occur are compile-time errors, and if they have to be runtime do you have adequate tests and documentation around those runtime errors? [For more details](https://github.com/wwt/Workflow/blob/main/CONTRIBUTING.md#errors).

----

### If Applicable:

----


WorkflowExample/WorkflowExample/Ordering/ReviewOrderViewController.swift
```diff
@@ -11,38 +11,39 @@ import Workflow
 import WorkflowUIKit
 import UIKit
 
-class ReviewOrderViewController: UIWorkflowItem<Order, Order?>, StoryboardLoadable {
-    var order: Order?
+class ReviewOrderViewController: UIWorkflowItem<Order, Order>, StoryboardLoadable {
```
Richard-Gist, line 14: As discussed, we'll want to have added documentation that `StoryboardLoadable` should be attached to the class instead of an extension to get the best experience (at this time).

Richard-Gist: As well as `FlowRepresentable` for those who want to programmatically create their view controllers.

Workflow/Sources/Workflow/Protocols/FlowRepresentable.swift
```diff
@@ -22,10 +22,11 @@ public protocol FlowRepresentable {
 
     var _workflowPointer: AnyFlowRepresentable? { get set }
     var _workflowUnderlyingInstance: Any { get }
-    /// instance: A method to return an instance of the `FlowRepresentable`
-    /// - Returns: `AnyFlowRepresentable`. Specifically a new instance from the static class passed to a `Workflow`
-    /// - Note: This needs to return a unique instance of your view. Whether programmatic or from the storyboard is irrelevant
-    static func instance() -> Self
```
Richard-Gist, line 28: This should be marked as `unavailable` and point consumers to how to upgrade.

Richard-Gist (changes requested):

Workflow/Sources/Workflow/Protocols/FlowRepresentable.swift
Tyler-Keith-Thompson: In a frustrating and shocking twist members can only be marked unavailable in objective-c protocols and it is completely impossible to represent a FlowRepresentable in Objective-C

WorkflowExample/WorkflowExample/Ordering/ReviewOrderViewController.swift
Tyler-Keith-Thompson: Added docs

Richard-Gist (approved):
Thanks for getting this together.  We'll finish documentation in #10

PR #26, opened 2021-05-24
Swiftpm support

Tyler-Keith-Thompson:
## Checklist:
- [ ] Is the [static analysis tool](https://sonarcloud.io/dashboard?id=wwt_Workflow) reporting 0 issues?

----

### If Applicable:

----


Tyler-Keith-Thompson:
Heya folks. Thought I'd save somebody a little trouble and move the swiftpm stuff up a directory so people can actually use swiftPM. Github actions is lying and saying the workspace does not contain the scheme "Workflow" but it most certainly does, and running `fastlane unit_test` locally works just fine.

I've had some issues today with the github macos runner for my other repos, so I'm hand waiving and saying it's lying to us? It could also have something to do with me being on big sur.

Tyler-Keith-Thompson:
https://community.sonarsource.com/t/sonarswift-code-coverage/17240/5

Something is wrong with our xcode coverage shell script but I haven't the faintest idea what, it is likely unrelated to the PR.

morganzellers (approved):

.github/workflows/CI.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       - name: Run TESTS
         run: bundle exec fastlane unit_test
       - name: Generate SonarCloud Friendly Coverage Report
-        run: ./xccov-to-sonarqube-generic.sh ~/Library/Developer/Xcode/DerivedData/*/Logs/Test/*.xcresult > sq-generic.xml
+        run: ./xccov-to-sonarqube-generic.sh "~/Library/Developer/Xcode/DerivedData/*/Logs/Test/*.xcresult" > "sq-generic.xml"
```
Richard-Gist, line 13: The internet lied to you.  These aren't the issue.

Workflow.xcworkspace/xcshareddata/xcschemes/Workflow.xcscheme
```diff
@@ -20,113 +20,43 @@
...
-            BlueprintName = "WorkflowUIKit"
-            ReferencedContainer = "container:">
-         </BuildableReference>
-      </CodeCoverageTargets>
-      <TestPlans>
-         <TestPlanReference
-            reference = "container:../Workflow.xctestplan"
-            default = "YES">
-         </TestPlanReference>
-      </TestPlans>
```
Richard-Gist, line 100: Looks like you dropped the test plan.  This is one of the reasons why coverage generation step is failing.  Essentially you changed the structure of the DerivedData folder to now be a single level instead of nested in a Workflow* folder.

Additionally when we found this, it seems that only 94 of the 225 tests are running.

docs/index.html
```diff
@@ -20,7 +20,7 @@
     <header class="header">
```
Richard-Gist, line 20: I don't know how to comment on the above BIN file.  What is the "docs/docsets/Workflow.tgz" file used for?

Workflow.xcworkspace/xcshareddata/xcschemes/Workflow.xcscheme
Richard-Gist: @morganzellers and I are looking into this issue and what the fix looks like,  but it will need to be resolved before merging in.

Richard-Gist (changes requested):

Workflow.xcworkspace/xcshareddata/xcschemes/Workflow.xcscheme
Tyler-Keith-Thompson: Test plan removal was a partially deliberate choice. Xcode 12.5 is getting much better about just opening Package.swift files and having it work. They're actually removing the xcodeproj generation, if you run it now you get a warning. 

Because of this it seems optimal to attempt to keep as close to what Xcode generates from the package.swift as is reasonably possible. That line of reasoning is what caused me to remove the test plan, it's not a hill I'll die on if we want to just bring it back.

docs/index.html
Tyler-Keith-Thompson: No clue, jazzy makes a thing, I check in the thing jazzy creates.

Workflow.xcworkspace/xcshareddata/xcschemes/Workflow.xcscheme
Richard-Gist: Also an important note, is that our pipeline is still running 12.4.  I'm not sure if that's MS lagging behind or if we have something specified.  Either way we'll have to keep that in consideration as we move forward.

Tyler-Keith-Thompson: Yeah there's a big thread on Github about that. The official answer was something like "We are working on it, if you want better you can pay MacStadium or use your own runner". Apparently Azure DevOps has the same challenges right now. I suspect the new chipset and general VM issues with Big Sur are making it hard for everybody right now.

What you should be careful of is that you specify macos-latest which means as soon as they update your builds will start using 12.5. It *may* be worth specifying a version to avoid breakage, but that also means you have to come up with your own cadence to review their hosted machine updates and choose to update.

Tyler-Keith-Thompson: https://github.com/actions/virtual-environments/issues/2486

Richard-Gist: Ok since, you intentionally removed the TestPlan, and it doesn't seem to impact my Xcode which is on 12.4, we are fine with this change.

.github/workflows/CI.yml
Richard-Gist: We've made an update to this path to point directly at the new location.  Unfortunately my `**/*.xcresult` did not work, so we still have to point at the correct directory.  Upon doing that, we now seem to be gather test coverage for all 226 tests in the run.

Unfortunately (again) we're still not getting the coverage all the way to Sonarcloud.  So we'll have to keep looking deeper.

Richard-Gist (approved):
Seems to be working now.  You will need to go update all the issues on Sonarcloud.  I do not seem to have that ability.  Be sure to bring over the reasons from before you moved the file.

PR #29, opened 2021-05-24
Removing need for the Info.plist for versioning

Richard-Gist:
…which was already deleted in another PR - RAG

## Checklist:
- [ ] Is the [static analysis tool](https://sonarcloud.io/dashboard?id=wwt_Workflow) reporting 0 issues?

----

### If Applicable:

----

### DIFF
--- a/.github/workflows/trunk.yml
+++ b/.github/workflows/trunk.yml
@@ -19,8 +19,6 @@ jobs:
         id: bump_version
         run: |
           bundle exec fastlane patch
-          VERSION=`/usr/libexec/PlistBuddy -c "Print :CFBundleVersion" Workflow/Info.plist`
-          echo "##[set-output name=version;]$VERSION"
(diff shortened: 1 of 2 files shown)


Richard-Gist:
The Info.plist was being used solely for updating the version of the library though not used when consuming the libraries.  This change fixes that need (also because Tyler already deleted it in the SwiftPM PR, and we all missed it).

So now we echo out the new version of the podspec directly into the environment variables from fastlane, instead of into a plist to then be read out into a variable to then be echoed into an environment variable.  I tested this on my own repo: [actions-testing](https://github.com/Richard-Gist/actions-testing/pull/56) to make sure that it would work, and it seems to be working as expected.

morganzellers (approved):

Richard-Gist:
I merged after a team discussion that we were good to merge.

PR #31, opened 2021-05-26
Fixes issue when abandoning a multiple modal workflow

morganzellers:
## Checklist:

----

### If Applicable:

----


morganzellers:
Updated abandon function to be less specific around conditions for modal dismissal. - MZ BL

WorkflowExample/WorkflowUIKitTests/UIKitConsumerAbandonTests.swift
```diff
@@ -219,6 +219,33 @@ class UIKitConsumerAbandonTests: XCTestCase {
...
+        XCTAssertUIViewControllerDisplayed(ofType: FR1.self)
+        XCTAssertNotNil(UIApplication.topViewController())
+        (UIApplication.topViewController() as? FR1)?.proceedInWorkflow(nil)
+        XCTAssertUIViewControllerDisplayed(ofType: FR2.self)
+        XCTAssertNil(UIApplication.topViewController()?.navigationController)
+        (UIApplication.topViewController() as? FR2)?.proceedInWorkflow(nil)
+        XCTAssertUIViewControllerDisplayed(ofType: FR3.self)
+        (UIApplication.topViewController() as? FR3)?.proceedInWorkflow(nil)
+        XCTAssertUIViewControllerDisplayed(ofType: FR4.self)
+        (UIApplication.topViewController() as? FR4)?.abandonWorkflow()
```
Richard-Gist, line 246: This is calling `abandonWorkflow()` on the `FlowRepresentable` but the issue calls out that it is about calling `abandon()` on the `Workflow`.  They ultimately end up in the same spot, but it's not exactly following the same code path as the bug.  I think this test is important to test what happens when a FlowRepresentable abandons the workflow in a chain of modals, but we need another test for when the workflow is abandoned.

Sources/WorkflowUIKit/UIKitPresenter.swift
```diff
@@ -84,8 +84,8 @@ open class UIKitPresenter: OrchestrationResponder {
                     nav.popViewController(animated: animated)
                 }
             }
-        } else {
-            first.dismiss(animated: animated, completion: onFinish)
+        } else if let presenting = first.presentingViewController {
```
Richard-Gist, line 87: It looks like this has an unhandled situation for when there is no navigation controller and the view does not have a presentingViewController.   Is that a scenario that users could get into?  I think it's possible, but let me know if that has already been explored and deemed impossible.

Richard-Gist (changes requested):

Sources/WorkflowUIKit/UIKitPresenter.swift
morganzellers: Re-added the last `else` condition in. @brianlombardo and I had thought the `else if` we added would cover all scenarios, but it looks like there are tests still covering the last `else`, so I added back.

WorkflowExample/WorkflowUIKitTests/UIKitConsumerAbandonTests.swift
morganzellers: Updated this test based on other requested fixes and added a new test for this requested fix

Sources/WorkflowUIKit/UIKitPresenter.swift
```diff
@@ -86,6 +86,8 @@ open class UIKitPresenter: OrchestrationResponder {
             }
         } else if let presenting = first.presentingViewController {
             presenting.dismiss(animated: animated, completion: onFinish)
```
Richard-Gist, line 88: I could be misremembering how dismiss works, but doesn't it dismiss the view you call it on?  But this seems to be dismissing up to the view we are calling it on.  That makes me wonder if there are some exploratory scenarios we should look at (not necessarily unit test) around chaining workflows in different scenarios.

Or you know, if y'all already did all of that, then we're good and you can tell me to stop.

Sources/WorkflowUIKit/UIKitPresenter.swift
Richard-Gist: There are tests covering them, but are they asserting on it?  I think tests that cover but not assert could be looked at for new tests that do assert.  Y'all have passed the pipeline with and without the `else` so I think we will still need to add tests around this.

Richard-Gist (changes requested):

Sources/WorkflowUIKit/UIKitPresenter.swift
wiemerm: We should check to see if having `ViewController` present another viewcontroller (I'll call this `ViewControllerB`) which in turn launches a workflow. Does the dismiss send us back to `ViewControllerB` or back to `ViewController` and which is the appropriate/expected behavior?

Richard-Gist: Sending back to `ViewControllerB` is what I think the correct behavior should be.  I've abandoned my Workflow so all things related to my workflow I closed, but not the thing that launched the workflow.

morganzellers: `dismiss` has separate functionalities depending on the calling view controller.  We are forcing it to always use the behavior where the calling view controller relies on the presenting view controller to dismiss.

Sources/WorkflowUIKit/UIKitPresenter.swift
morganzellers: Xcode was reporting coverage of the else, but we were unable to hit breakpoints, print(), or fatalError() functions inside of the else. As such, we removed the untestable line.

Richard-Gist (approved):
Thank you for all the hard work!

PR #33, opened 2021-06-02
Fixes removedAfterProceeding not dismissing final view when applied to final item in Workflow.

### LINKED ISSUE #19: removedAfterProceeding on last item in Workflow
### Describe the bug
If you have `.removedAfterProceeding` as the `FlowPersistence` on an item in a workflow it does not properly remove that item (UIKit)

## To Reproduce
Create 3 `FlowRepresentable`s the last of which has a `flowPersistence` of `.removedAfterProceeding`
Call `proceedInWorkflow()` on all items in that flow
Observe after OnFinish is called the final `UIViewController` is still displayed

### Expected behavior
The `UIViewController` should not be displayed after the onFinish block has executed

### Screenshots
NO

Richard-Gist:
## Checklist:
- [X] Do we need to [increment the minor/major version](https://github.com/wwt/Workflow/blob/main/CONTRIBUTING.md#versioning)?
- [X] Did you [change the public API](https://github.com/wwt/Workflow/blob/main/CONTRIBUTING.md#public-api)?

----

### If Applicable:

----


Richard-Gist:
This should have been a draft!!

EOD:  We have decided to create a "complete" function on OrchestrationResponder and it will have the sole responsibility of calling the onFinish provided to the Workflow.  In driving out these tests, we decided that the `MockOrchestrationResponder` should have the ability to call a default implementation. In this case, the default is that complete will call onFinish with the passedArgs.

These decisions are being worked through and the current state of the PR is incomplete for all of those decisions.

Sources/WorkflowUIKit/UIKitPresenter.swift
```diff
@@ -61,6 +61,19 @@ open class UIKitPresenter: OrchestrationResponder {
         abandon(workflow, animated: true, onFinish: onFinish)
     }
 
+    public func complete(_ workflow: AnyWorkflow, passedArgs: AnyWorkflow.PassedArgs, onFinish: ((AnyWorkflow.PassedArgs) -> Void)?) {
+        // AnyWorkflow does not have a last(where:)
+        // swiftlint:disable:next last_where
+        let lastInstance = workflow.filter { $0.value.instance != nil }.last
```
Tyler-Keith-Thompson, line 67: It wouldn't be hard to add.... Workflow has a last(where:) and you'd just have to call it, exactly like we did to make AnyWorkflow conform to `Sequence`

Sources/Workflow/Models/Workflow.swift
```diff
@@ -191,7 +192,8 @@ public final class Workflow<F: FlowRepresentable>: LinkedList<_WorkflowItem> {
         }
 
         guard let nextNode = nextLoadedNode else {
-            onFinish?(argsToPass)
+            #warning("Workflow in Complete is incorrect. Test Drive it to make it correct.")
+            orchestrationResponder?.complete(AnyWorkflow(Workflow(self.last)), passedArgs: argsToPass, onFinish: onFinish)
```
Richard-Gist, line 196: Ok so this is a little odd.  Because I used a node from the linked list the first time, we ended up with a convoluted `AnyWorkflow(self)` which is correct.  Then this code is now actually wrong, BUT it won't fail tests because it still falls under what we are testing (the last item is persisted or not).

So we're still kind of right but really we either need to add tests or finally set it completely right with:
```suggestion
            orchestrationResponder?.complete(AnyWorkflow(self), passedArgs: argsToPass, onFinish: onFinish)
```

Sources/WorkflowUIKit/UIKitPresenter.swift
wiemerm: Did the thing. Happy?

Tyler-Keith-Thompson: Quite! 🎉

Sources/Workflow/Models/Workflow.swift
wiemerm: We added a test.

wiemerm:
EOD: We addressed existing comments.

morganzellers (approved):

wiemerm:
EOD:  We continued to drive out changes to the destroy function and handling the onFinish() based on earlier decisions made.

morganzellers (approved):

Tests/WorkflowExampleTests/Mocks/MockOrchestrationResponder.swift
```diff
@@ -10,6 +10,10 @@ import Foundation
 import Workflow
 
 class MockOrchestrationResponder: OrchestrationResponder {
+    func complete(_ workflow: AnyWorkflow, passedArgs: AnyWorkflow.PassedArgs, onFinish: ((AnyWorkflow.PassedArgs) -> Void)?) {
+
```
Richard-Gist, line 14: Needs to be implemented.  Probably should be copied from the other MockOrchestrationResponder.

wiemerm, line 14: Needs implementation

Tests/WorkflowTests/LinkedListTests.swift
```diff
@@ -409,6 +409,34 @@ class LinkedListTests: XCTestCase {
         XCTAssertEqual(list.last?.value, 4)
     }
 
+
```
wiemerm, line 412: Delete this line; no reason for 2 empty lines here

Richard-Gist (commented):
Also, we need to generate new docs for this.

wiemerm (changes requested):

Sources/Workflow/Models/Workflow.swift
```diff
@@ -123,7 +123,7 @@ public final class Workflow<F: FlowRepresentable>: LinkedList<_WorkflowItem> {
         }
 
         guard let first = firstLoadedInstance  else {
-            onFinish?(passedArgs)
+            orchestrationResponder.complete(AnyWorkflow(self), passedArgs: passedArgs, onFinish: onFinish)
```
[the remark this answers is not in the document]
wiemerm: We do have a test that fails by asserting that the complete was called. We could probably add a test to exercise UIKit in this scenario.

Sources/Workflow/Protocols/OrchestrationResponder.swift
```diff
@@ -32,6 +32,13 @@ public protocol OrchestrationResponder {
      - Parameter onFinish: A closure that is executed when the responder is finished abandoning.
      */
     func abandon(_ workflow: AnyWorkflow, onFinish: (() -> Void)?)
+    /**
+     Respond to the `Workflow` completing.
+     - Parameter workflow: The `AnyWorkflow` that is being completed.
+     - Parameter passedArgs: The `AnyWorkflow.PassedArgs` to be passed to `onFinish`.
+     - Parameter onFinish: A closure that is executed when the responder is finished completing.
+     */
+    func complete(_ workflow: AnyWorkflow, passedArgs: AnyWorkflow.PassedArgs, onFinish: ((AnyWorkflow.PassedArgs) -> Void)?)
```
Richard-Gist, line 41: I don't see it as a conversation to resolve so I'm adding here.
1. We need to generate the jazzy docs for this change
2. We need to version ramp the minor version of the app as we are adding Complete to the API.

morganzellers: I see Release CI has been bumping the patch version on merges to main, how do we increment the minor spot?

Sources/Workflow/Models/Workflow.swift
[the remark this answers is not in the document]
Richard-Gist: I have added a test that will cover this for UIKitPresenter.

Sources/Workflow/Protocols/OrchestrationResponder.swift
Richard-Gist: I have version bumped the minor with `bundle exec fastlane minor`

Sources/WorkflowUIKit/UIKitPresenter.swift
```diff
@@ -89,12 +101,21 @@ open class UIKitPresenter: OrchestrationResponder {
...
+    private func destroy(_ view: UIViewController, completion: (() -> Void)? = nil) {
         if let nav = view.navigationController {
             let vcs = nav.viewControllers.filter {
                 $0 !== view
             }
-            nav.setViewControllers(vcs, animated: false)
+
+            if vcs.isEmpty && nav === internalNavigationController {
+                nav.presentingViewController?.dismiss(animated: false) { [weak self] in
+                    self?.internalNavigationController?.setViewControllers(vcs, animated: false)
```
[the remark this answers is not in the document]
Richard-Gist: Yes this line needs it and I don't know how to test..... I'll share with the class tomorrow.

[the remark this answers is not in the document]
Richard-Gist: Actually, I'm full of lies and that line is not needed and I was just needing to give it more time to cleanup.  No leak here.

wiemerm (dismissed):

wiemerm:
Good to merge after the docs are updated.

Sources/Workflow/Protocols/OrchestrationResponder.swift
Richard-Gist: Last commit generated the docs and added some more.

wiemerm (approved):

morganzellers (approved):

PR #37, opened 2021-06-10
Creates Getting Started with Programmatic UIKit

wiemerm:
## Checklist:

----

### If Applicable:

----

### DIFF
--- a/wiki/Getting-Started-UIKit.md
+++ b/wiki/Getting-Started-UIKit.md
@@ -90,12 +90,12 @@ class SecondViewController: UIWorkflowItem<String, String>, FlowRepresentable {
 
     required init?(coder: NSCoder) { nil }
 
-    @objc private func finishPressed() {
-        proceedInWorkflow(email)
-    }
-
     func shouldLoad() -> Bool {
         return email.contains("@wwt.com")
+    }
+
+    @objc private func finishPressed() {
+        proceedInWorkflow(email)
     }
 }
 ```
@@ -116,11 +116,10 @@ import Workflow
 class ViewController: UIViewController {
     private let launchButton = UIButton()
 
-    init() {
-        super.init(nibName: nil, bundle: nil)
-
+    override func viewDidLoad() {
         launchButton.setTitle("Launch Workflow", for: .normal)
-        launchButton.addTarget(self, action: #selector(launchWorkflow), for: .touchUpInside)
+        launchButton.setTitleColor(.systemBlue, for: .normal)
+        launchButton.addTarget(self, action: #selector(didTapLaunchWorkflow), for: .touchUpInside)
 
         view.addSubview(launchButton)
 
@@ -129,11 +128,7 @@ class ViewController: UIViewController {
         launchButton.centerYAnchor.constraint(equalTo: view.centerYAnchor).isActive = true
     }
 
-    required init?(coder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
-    }
-
-    @objc private func launchWorkflow() {
+    @objc private func didTapLaunchWorkflow() {
         let workflow = Workflow(FirstViewController.self)
             .thenPresent(SecondViewController.self)
 
--- a/wiki/Getting-Started-UIKit.md
+++ b/wiki/Getting-Started-UIKit.md
@@ -17,7 +17,6 @@ Create two view controllers that inherit from [UIWorkflowItem<I, O>](https://git
 ```swift
 import UIKit
 import Workflow
-import WorkflowUIKit
 
 class FirstViewController: UIWorkflowItem<String, String>, FlowRepresentable {
     private let name: String
@@ -69,8 +68,7 @@ class FirstViewController: UIWorkflowItem<String, String>, FlowRepresentable {
 // This screen shows an employee only screen
 class SecondViewController: UIWorkflowItem<String, String>, FlowRepresentable {
     private let email: String
-    private let emailLabel = UILabel()
-    private let nextButton = UIButton()
+    private let finishButton = UIButton()
 
     required init(with email: String) {
         self.email = email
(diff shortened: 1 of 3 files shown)


wiemerm:
EOD: We created a sample project to build out what Getting Started with Programmatic views and added an initial version of what this Getting Started would look like using the Getting Started with Storyboards as a reference.

We discussed whether to break out the view code into separate views or keep them within the UIViewControllers. We chose to put them in the view controllers in order to cut down on boilerplate code and avoid prescribing a particular way of setting up views.

wiki/Getting-Started-UIKit.md
```diff
@@ -0,0 +1,248 @@
+# Interested in seeing examples of Workflow in action?
+Start by cloning the repo and checking out the 'WorkflowExample' scheme. This should give you a decent idea of how the library works.  If you want to create a new project, read on.
+
+# CocoaPods with Programmatic Views
+
+## Adding the dependency
```
Tyler-Keith-Thompson, line 6: We should have a conversation about what our preferred package manager is. Most of our instructions reference CocoaPods but it seems maybe more likely our preferred package manager should be SwiftPM.

wiki/Getting-Started-UIKit.md
```diff
@@ -0,0 +1,248 @@
...
+        emailTextField.widthAnchor.constraint(equalToConstant: 300).isActive = true
+
+        saveButton.translatesAutoresizingMaskIntoConstraints = false
+        saveButton.centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
+        saveButton.topAnchor.constraint(equalTo: emailTextField.bottomAnchor, constant: 24).isActive = true
+    }
+
+    required init?(coder: NSCoder) { nil }
+
+    @objc
```
Tyler-Keith-Thompson, line 64: This violates our styleguide. I weirdly cannot find a link to the exact rule but annotations go on the same line as the function declaration, they aren't broken apart above it. You can see several examples if you search the styleguide for `@`. This fits with SwiftUI conventions like how you have
```swift
// RIGHT
@State var thing: String = ""

// WRONG
@State
var thing: String = ""
```

wiemerm: I could not find a specific rule for this either, but I was ultimately able to find it played out in some of the examples for other rules. Based on this, I think we should consider adding this as a rule in the styleguide so that it is more explicit.

Richard-Gist:
> We discussed whether to break out the view code into separate views or keep them within the UIViewControllers. We chose to put them in the view controllers in order to cut down on boilerplate code and avoid prescribing a particular way of setting up views.

I've been out of the programmatic views for awhile.  If this is what we would do for our projects, we're good.  If we would not do it for our projects, we should have a call out that we are doing it for the sake of the guide.

wiki/Getting-Started-UIKit.md
Richard-Gist: Discussion started: https://github.com/wwt/Workflow/discussions/40

wiki/Getting-Started-UIKit.md
```diff
@@ -0,0 +1,239 @@
...
+
+        view.addSubview(launchButton)
+
+        launchButton.translatesAutoresizingMaskIntoConstraints = false
+        launchButton.centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
+        launchButton.centerYAnchor.constraint(equalTo: view.centerYAnchor).isActive = true
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
Richard-Gist, line 133: Completely wiping out the ViewController from the storyboards may be going too far with Programmatic?  I'm not sure, but initially on copy pasting my way through the tutorial this fatalError is thrown because this view is no longer loadable from a storyboard.

I suppose this is dependent on if people who do programmatic fully configuring their projects every time to be without the storyboard or if they simple leverage the existing storyboard starting view and then customize from there.

wiki/Getting-Started-UIKit.md
```diff
@@ -0,0 +1,239 @@
...
+Next, we create a [Workflow](https://gitcdn.link/cdn/wwt/Workflow/faf9273f154954848bf6b6d5c592a7f0740ef53a/docs/Classes/Workflow.html) that is initialized with our [FlowRepresentable](https://gitcdn.link/cdn/wwt/Workflow/faf9273f154954848bf6b6d5c592a7f0740ef53a/docs/Protocols/FlowRepresentable.html)s and launch it from a view controller that is already loaded onto the screen (in our case, the 'ViewController' class provided by Xcode).
+
+```swift
+import UIKit
+import Workflow
+
+class ViewController: UIViewController {
+    private let launchButton = UIButton()
+
+    init() {
```
Richard-Gist, line 119: would it make sense to change this to the viewDidLoad?

Richard-Gist (commented):
Not done reviewing but hit enough issues to share my comments.

wiki/Getting-Started-UIKit.md
wiemerm: In order to allow it to be copied and pasted regardless of whether an individual kept storyboards or fully replaced storyboards then yet, it makes sense to switch that to viewDidLoad().

wiki/Home.md
```diff
@@ -1,4 +1,5 @@
 - [Why This Library?](https://github.com/wwt/Workflow/wiki/Why-This-Library%3F)
 - [Installation](https://github.com/wwt/Workflow/wiki/Installation)
-- [Getting Started](https://github.com/wwt/Workflow/wiki/getting-started)
+- [Getting Started with Storyboards](https://github.com/wwt/Workflow/wiki/getting-started)
+- [Getting Started with Programmatic UIKit](https://github.com/wwt/Workflow/wiki/getting-started-uikit)
```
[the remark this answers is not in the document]
wiemerm: Added.

wiki/Getting-Started-UIKit.md
Richard-Gist: As this is a bigger discussion to have and there is an existing discussion linked.  I'm resolving this conversation, and it should continue in #40

wiki/Getting-Started-UIKit.md
Richard-Gist: Issue tracked with: #42

Richard-Gist (approved):

PR #39, opened 2021-06-11
Fixes the underlying LinkedList retain cycle issue

### LINKED ISSUE #38: Memory Leak on Workflows
### Describe the bug
We're leaking memory every time a workflow is launched and followed. 

## To Reproduce
1. Create a workflow (ideally that autocompletes) for example a workflow that proceeds to the next item on `viewDidLoad` after 1 second.
2. Launch that workflow multiple times
3. Make sure the workflow has called `onFinish`. Bonus points: call `abandon` on it, NOTE: You should not have to in order to solve this problem, and calling `abandon` exposed an *additional* bug.
3. Open the memory graph and cry, because the `_WorkflowItem`s got retained when they should not have.

### Expected behavior
A workflow should not create retain cycles, it's fine that it retains while executing, but when it is abandoned or when the same workflow is launched again it is reasonable that there aren't retain cycles.

NOTE: The library clears instances when you `launch` a Workflow. Is that behavior correct, or is that a bad assumption? It probably has limitations if a user launches the *same* instance of a workflow multiple times and expects it to just be duplicated. That's an odd thing for a user to expect, but the library does not support it currently. It assumes every call to `launch` means it can clear memory of all known instances.

Tyler-Keith-Thompson:
…Had to comment out sort tests because something very strange is happening with it - TT

## Checklist:

----

### If Applicable:

----


Tyler-Keith-Thompson:
I have functioning code that resolves the memory leak, but it has caused unexpected behavior with `sort`. I thought I'd create a draft pull request in case the team decided that sorting was sort of a silly operation to even allow on a `Workflow` anyways and wanted to just delete the sort code and move on. 

I will continue to work on fixing that in my spare time.

For those curious where the memory leak was, it's just that the `LinkedList` needed a `deinit` to clean up strong references on nodes. Nothing major, I've got a test around that.

Tyler-Keith-Thompson:
I withdraw my last statement, that sorting issue exposed a much more serious one with how `removeTillEnd` worked. Can't believe that was missed. It's now all fixed, and we're much better about memory usage AND we don't have an infinitely looping function in the codebase anymore. Yay!

Sources/Workflow/LinkedList/LinkedListComparable.swift
```diff
@@ -13,12 +13,14 @@ extension LinkedList where Value: Comparable {
     /// Sorts the linked list in place using a merge sort.
```
Richard-Gist, line 13: Isn't this a lie now, and you are not sorting in place?  If I have a reference to a node, and I call sort on the linked list, my node is not sorted, but my list is.  All references to my nodes are outdated at this point, right?

It could more realistically call out that it mutates the linked list into a sorted list, with an important note that previous references to nodes are now outdated.

Sources/Workflow/Models/Workflow.swift
```diff
@@ -286,7 +286,7 @@ extension Workflow where F.WorkflowOutput == Never {
     public func thenProceed<FR: FlowRepresentable>(with type: FR.Type,
                                                    launchStyle: LaunchStyle = .default,
                                                    flowPersistence: @escaping @autoclosure () -> FlowPersistence = .default) -> Workflow<FR> where FR.WorkflowInput == Never {
-        let wf = Workflow<FR>(first)
+        let wf = Workflow<FR>(first?.copy())
```
Richard-Gist, line 289: This `.copy()` seems to be the crux of the memory leak and the most important part of these `thenProceed`s.  Would it make sense to extract it as a preferred method for creating a fluent describer(?) of the workflow?

Spitballing, but something like:
```suggestion
        let wf = Workflow<FR>(fluently: first)
```

Sources/Workflow/LinkedList/LinkedListMutatingOperations.swift
```diff
@@ -218,7 +218,9 @@ extension LinkedList {
      - Important: This operation mutates the original `LinkedList`.
```
Richard-Gist, line 218: Assuming my earlier comment is correct, we should call out that it does *not* mutate the nodes.

Sources/Workflow/Models/Workflow.swift
Tyler-Keith-Thompson: Hrm, I think the `copy()` thing wasn't clear to you. The problem is that if you pass a direct reference to `first` with the new `deinit` it just gets cleaned up, so like all the proceeds don't work. 

I am unclear what `fluently` means here? Especially since it's not fluent? Another potential fix would be in the `LinkedList` intializer to make it auto-copy, rather than telling each calling instance of it to copy.

Sources/Workflow/LinkedList/LinkedListComparable.swift
Tyler-Keith-Thompson: Perhaps, it's not a lie that it's sorted in place, because the *LinkedList* still is. But I agree I need to call out any references to nodes won't be changing. Lemme freak out over docs on that for a bit.

Sources/Workflow/LinkedList/LinkedListMutatingOperations.swift
Tyler-Keith-Thompson: done

Sources/Workflow/Models/Workflow.swift
Tyler-Keith-Thompson: I think you'll find my solution satisfactory

Sources/Workflow/LinkedList/LinkedList.swift
```diff
@@ -53,8 +53,15 @@ public class LinkedList<Value>: Sequence, CustomStringConvertible {
...
-    /// Creates a `LinkedList` by providing the first node in the list.
+    /// Creates a copy of a `LinkedList` by providing the first node, and copying it.
     public required init(_ node: Element? = nil) {
+        first = node?.copy()
+    }
+
+    /** Creates a `LinkedList` by providing the first node in the list.
+    - Important: This can potentially cause memory retention issues, you are passing a reference, be aware.
+     */
+    public required init(withoutCopying node: Element?) {
```
Richard-Gist, line 64: Nice I like this even more than my suggestion.

Richard-Gist: We need to generate the docs for these and other doc changes.  Would you like me to do that before merging or do you want to do it?

Tyler-Keith-Thompson: It’s all you my friend

Richard-Gist (approved):

PR #63, opened 2021-06-24
Adds preference for attributes on the same line as what they attribute to the StyleGuide

wiemerm:
## Checklist:

----

### If Applicable:

----

### DIFF
--- a/.github/STYLEGUIDE.md
+++ b/.github/STYLEGUIDE.md
@@ -743,13 +743,13 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
   ```
   </details>
 
-* **PREFER attributes on the same line as what they are attributing.**
+* **PREFER attributes on the same line as what they are attributing for properties, functions, and closures.**
 
   <details>
 
   #### Why?
 
-    Because we naturally read lines of code as a consistent line, introducing whitespace makes it that much harder to understand the full context of the line you're trying to understand. This is the same reasoning used for preferring a return on the same line as the final parameter in a function declaration.
+    Because we naturally read lines of code as a consistent line, introducing whitespace makes it that much harder to understand the full context of the line you're trying to understand. This is the same reasoning used for preferring a return on the same line as the final parameter in a function declaration. A notable exception to this preference is the `@available` attribute which should always be on its own line.
 
     NOTE: This does not come at the cost of sanity. If you have an exceptionally large number of attributions then it might indicate that you've gotten overzealous, and it actually obscures your meaning, rather than add to it.
 
@@ -765,10 +765,60 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
     // ...
   }
 
+  // WRONG
+  @available(iOS 14.0, *) func doSomething() {
+    // ...
+  }
+
   // RIGHT
   @State var stateVar: Bool = true
   @discardableResult func thing() -> Bool { true }
-  @MainActor class MyClass { 
+  @MainActor @ViewBuilder func createViews() {
+    // ...
+  }
+
+  // RIGHT
+  @available(iOS 14.0, *) 
+  func doSomething() {
+    // ...
+  }
+  ```
+  </details>
+
+* **PREFER attributes on a separate line as what they are attributing for classes, structs, and actors.**
+
+  <details>
+
+  #### Why?
+  This is the convention set by Apple in the [Swift attribute docs](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html). A notable exception to this preference is the `@available` attribute which should always be on its own line.
+    
+
+  ```swift
+  // WRONG
+  @dynamicCallable class MyClass {
+    // ...
+  }
+
+  // WRONG
+  @MainActor struct MyStruct {
+    // ...
+  }
+
+  // WRONG
+  @available(iOS 14.0, *) @MainActor 
+  struct MyStruct {
+    // ...
+  }
+  
+  // RIGHT
+  @dynamicCallable class MyClass {
+    // ...
+  }
+
+  // RIGHT
+  @available(iOS 14.0, *) 
+  @propertyWrapper 
+  struct MyPropertyWrapper {
     // ...
   }
   ```
--- a/.github/STYLEGUIDE.md
+++ b/.github/STYLEGUIDE.md
@@ -785,7 +785,7 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
   ```
   </details>
 
-* **PREFER attributes on a separate line as what they are attributing for classes, structs, and actors.**
+* **PREFER attributes on a separate line from what they are attributing for classes, structs, and actors.**
 
   <details>
 
@@ -819,6 +819,13 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
   @available(iOS 14.0, *) 
   @propertyWrapper 
   struct MyPropertyWrapper {
+    // ...
+  }
+
+  // RIGHT
+  @available(iOS 13, *)
+  @resultBuilder @requires_stored_property_inits
+  class WorkflowBuilder {
     // ...
   }
   ```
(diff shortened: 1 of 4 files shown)


Tyler-Keith-Thompson:
Should we call out the `@available` attribute?

.github/STYLEGUIDE.md
```diff
@@ -743,6 +743,37 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
   ```
   </details>
 
+* **PREFER attributes on the same line as what they are attributing.**
```
Richard-Gist, line 746: Is this the rule that should also dictate order of attributes?  Is that a rule we need?
```swift
@MainActor @ViewBuilder func createViews() {
```
or
```swift
@ViewBuilder @MainActor func createViews() {
```

Richard-Gist (commented):
I had questions...

.github/STYLEGUIDE.md
```diff
@@ -743,6 +743,37 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
...
+
+  ```swift
+  // WRONG
+  @objc
+  func thingExposedToObjectiveC() {
+    // ...
+  }
+  
+  // WRONG
+  @ViewBuilder @discardableResult @MainActor @inlinable func betYouDoNotEvenKnowWhatThisFunctionDoesAnymore() {
```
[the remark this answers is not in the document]
wiemerm: This one should be clearer that what we mean is "You're doing something wrong if you have this many attributes" rather than "Split them up on separate lines"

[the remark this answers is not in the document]
Tyler-Keith-Thompson: Part of me wants to say that many attributes on a function or property are simply a smell and since this is a "PREFER" I'm tempted to say don't worry about it?

That being said I am not at all a fan of 1 attribute per line, I much prefer the logical breakdown of related attributes, if possible.

.github/STYLEGUIDE.md
Tyler-Keith-Thompson: The order of attributes should be the same kind of rules that we have for other ordering, based off of effect on the overall codebase. So `@MainActor` affects every call site, `@discardableResult` does as well. `@ViewBuilder` only affects that function really, and `@inlineable` is a somewhat transparent compiler optimization.

.github/STYLEGUIDE.md
```diff
@@ -743,6 +743,37 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
   ```
   </details>
 
+* **PREFER attributes on the same line as what they are attributing.**
+
+  <details>
+
+  #### Why?
+
+    Because we naturally read lines of code as a consistent line, introducing whitespace makes it that much harder to understand the full context of the line you're trying to understand. This is the same reasoning we put behind preferring a return on the same line as the final parameter in a function declaration.
```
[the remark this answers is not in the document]
Tyler-Keith-Thompson: "applied to" is fine. I am also not sure I like yours better, fun fact...

.github/STYLEGUIDE.md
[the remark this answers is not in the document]
Richard-Gist: I think I'm looking for an example of what is something that is readable but still long.  And maybe a why...

```swift
@DependencyInjected @objc @CustomWorkflowAttribute var variableName: String?
```

.github/STYLEGUIDE.md
```diff
@@ -743,6 +743,37 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
...
+
+    NOTE: This does not come at the cost of sanity. If you have an exceptionally large number of attributions then it might indicate that you've gotten overzealous, and it actually obscures your meaning, rather than add to it.
```
Tyler-Keith-Thompson, line 754: Should we call out the `@available` attribute?

wiemerm: Calling out splitting into 2 potential guidelines one for inline when attributes are on functions, properties, or closures and another for attribute on the line above for classes, structs, and actors.

`@available` is still an outlier in that it always goes on its own line, so yes `@available` makes sense to be called out specifically at least in regards to inline preference.

.github/STYLEGUIDE.md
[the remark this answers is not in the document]
wiemerm:
```suggestion
    Because we naturally read lines of code as a consistent line, introducing whitespace makes it that much harder to understand the full context of the line you're trying to understand. This is the same reasoning used for preferring a return on the same line as the final parameter in a function declaration.
```

Not sure I like this one any better than "applied to"

.github/STYLEGUIDE.md
Richard-Gist: I think 2 rules make sense.  We should also show an example of what they look like with multiple attributes so that I know if I'm inlining on a separate line or stacking them.  And I think `@available` is an exception entirely and is always on its own line by itself.

.github/STYLEGUIDE.md
Richard-Gist: I think this makes sense.  +1 to Same rules as our file organization

.github/STYLEGUIDE.md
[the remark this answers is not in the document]
Richard-Gist: If we keep this line for splitting into 2 rules, I think this one reads better.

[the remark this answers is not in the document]
morganzellers: I like `used for` here

.github/STYLEGUIDE.md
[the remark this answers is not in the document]
wiemerm: Added an example with 2 attributes on the same line.

.github/STYLEGUIDE.md
wiemerm: `@MainActor @ViewBuilder func createViews() {` was added as an example. Should the ordering preference be a part of the existing preferences or are we proposing a 3rd addition to call out order of attributes?

.github/STYLEGUIDE.md
```diff
@@ -743,6 +743,94 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
...
+
+  // WRONG
+  @available(iOS 14.0, *) func doSomething() {
+    // ...
+  }
+
+  // RIGHT
+  @State var stateVar: Bool = true
+  @discardableResult func thing() -> Bool { true }
+  @MainActor @ViewBuilder func createViews() {
```
Richard-Gist, line 776:
```suggestion
  @MainActor @ViewBuilder func createViews() -> some View {
```

.github/STYLEGUIDE.md
```diff
@@ -743,6 +743,94 @@ Each guide is broken into a few sections. Sections contain a list of guidelines.
...
+  }
+
+  // WRONG
+  @available(iOS 14.0, *) @MainActor 
+  struct MyStruct {
+    // ...
+  }
+  
+  // RIGHT
+  @dynamicCallable class MyClass {
```
Richard-Gist, line 814:
```suggestion
  @dynamicCallable
  class MyClass {
```

.github/STYLEGUIDE.md
Richard-Gist: I think a 3rd rule is needed for ordering.  It essentially repeats or summarizes the rule for file organization.  The rule holds true despite being in the stacked or inline format.  Items with the most impact come first.

wiemerm: This might be covered by the "Why" of "DO place content in the correct order..." but if we need more clarity I'm on board with either expanding the existing or creating a new "DO" for ordering attributes.

Richard-Gist: Expanding the existing one would be enough.  Just something to point at that says "Do it this way"

wiemerm:
EOD: Decided to make 2 rules, one for inlining attributes for properties, functions, and closures and another for a separate line for classes, structs, and actors. Currently working through expanding content order to be clearer that attributes are included.

Richard-Gist (approved):

.github/STYLEGUIDE.md
wiemerm: Expanded the examples of the correct ordering guideline.

brianlombardo (approved):

PR #72, opened 2021-07-13
BETA RELEASE - SwiftUI support

Tyler-Keith-Thompson:
## Checklist:
- [x] Do we need to [increment the minor/major version](https://github.com/wwt/SwiftCurrent/blob/main/.github/CONTRIBUTING.md#versioning)?

----

### If Applicable:

----


Richard-Gist:
EOD: We got just about all of this in place. Everything is tested that can be tested, and that which is not covered is not testable.  The last bit to do is protecting in a double launch situation.  There are 2 warnings left: double launch and a bundle identifier warning.

Everything is set up as a BETA product and we should be able to go to trunk soon.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
```diff
@@ -0,0 +1,215 @@
...
+        onAbandon.append(closure)
+        return WorkflowView(isPresented: $isPresented,
+                            workflow: workflow,
+                            onFinish: onFinish,
+                            onAbandon: onAbandon,
+                            passedArgs: passedArgs)
+    }
+}
+
+// MARK: thenProceeds
```
Richard-Gist, line 131: We can leave this if it is helpful but otherwise I think this was accidentally left in.
```suggestion
```

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
```diff
@@ -0,0 +1,215 @@
...
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+extension WorkflowView where Args == Never {
+    /**
+     Adds an item to the workflow; enforces the `FlowRepresentable.WorkflowOutput` of the previous item matches the args that will be passed forward.
+     - Parameter workflowItem: a `WorkflowItem` that holds onto the next `FlowRepresentable` in the workflow.
+     - Returns: a new `WorkflowView` with the additional `FlowRepresentable` item.
+     */
+    public func thenProceed<FR: FlowRepresentable & View>(with item: WorkflowItem<FR>) -> WorkflowView<FR.WorkflowOutput> where FR.WorkflowInput == Never {
+        var workflow = self.workflow
+        if workflow == nil {
```
Richard-Gist, line 141: We should reverse this `if` statement so that appending is the primary path and initialization is the secondary.  Reasoning: because we expect this `thenProceed` to be appending more than it is initializing.

This of course applies to all the typed versions of this.

Sources/SwiftCurrent_SwiftUI/WorkflowItem.swift
```diff
@@ -0,0 +1,114 @@
...
+        modifierClosure = {
+            // We are essentially casting this to itself, that cannot fail. (Famous last words)
+            // swiftlint:disable:next force_cast
+            let instance = $0.underlyingInstance as! F
+            $0.changeUnderlyingView(to: closure(instance))
+        }
+        return self
+    }
+
+    func factory(args: AnyWorkflow.PassedArgs) -> AnyFlowRepresentable {
```
Richard-Gist, line 61: could this be private?
```suggestion
    private func factory(args: AnyWorkflow.PassedArgs) -> AnyFlowRepresentable {
```

Tests/SwiftCurrent_SwiftUITests/PersistenceTests.swift
```diff
@@ -0,0 +1,513 @@
...
+
+import SwiftCurrent
+@testable import SwiftCurrent_SwiftUI // testable sadly needed for inspection.inspect to work
+
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+final class PersistenceTests: XCTestCase {
+    // MARK: RemovedAfterProceedingTests
+    func testRemovedAfterProceeding_OnFirstItemInAWorkflow() throws {
+        struct FR1: View, FlowRepresentable, Inspectable {
+            var _workflowPointer: AnyFlowRepresentable?
```
Richard-Gist, line 21: Because of where we are at at this point, I wonder if it is safe to change these to weak to prevent the retention cycle or if that could throw our tests into an unknown state?
```suggestion
            weak var _workflowPointer: AnyFlowRepresentable?
```

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -0,0 +1,376 @@
...
+                .thenProceed(with: WorkflowItem(FR1.self)
+                                .applyModifiers { $0.customModifier().background(Color.blue) })).inspection.inspect { viewUnderTest in
+            XCTAssertNoThrow(try viewUnderTest.vStack().callOnDisappear())
+            XCTAssertNoThrow(try viewUnderTest.vStack().anyView(0).emptyView())
+        }
+
+        wait(for: [expectViewLoaded], timeout: 0.3)
+    }
+
+    func testWorkflowOnlyLaunchesOnce() throws {
```
Richard-Gist, line 352: It's a slight thing to think about, this one makes sense.  We should only launch the workflow once when the view is visible so that they don't lose their place... but if the view disappears then re-appears (toggling isPresented) then the view will need to restart, correct?

We may need a test for that scenario and maybe some quick research on scrolling a view out of view and back.

Richard-Gist (commented):
First pass thoughts.  I still need to go through again inspecting documentation and again to review all the code (I only reviewed the code since the last changes I had seen).

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
Tyler-Keith-Thompson: Not sure I entirely agree. The way it reads now is based on what will happen chronologically, that makes more sense in my head. It starts as nil, then we must assign it, then it is not nil anymore.

SwiftCurrent.podspec
```diff
@@ -26,6 +26,12 @@ Pod::Spec.new do |s|
     ss.dependency 'SwiftCurrent/Core'
   end
 
+  s.subspec 'BETA_SwiftUI' do |ss|
+    ss.ios.deployment_target = '11.0'
```
wiemerm, line 30: Curious: Why this is 11.0 when everything we're doing for SwiftUI is targeting at least 14.0?

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
```diff
@@ -0,0 +1,215 @@
...
+///         }
+///     })
+///     .onAbandon { print("presentingWorkflowView is now false") }
+///     .onFinish { args in print("Finished 1: \(args)") }
+///     .onFinish { print("Finished 2: \($0)") }
+///     .background(Color.green)
+///  ```
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+public struct WorkflowView<Args>: View {
+    @Binding public var isPresented: Bool
```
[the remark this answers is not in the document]
wiemerm: Yes, at least internal would be preferable. There's no need for the consumer to be directly altering this `isPresented`.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
Richard-Gist: I was thinking in regards to the styleguide: [DO write if/else statements starting with the happy path.](https://github.com/wwt/SwiftCurrent/blob/main/.github/STYLEGUIDE.md#control-flow).  The happy path in this regard being appending, not initializing.  It would be something like:
```swift
if workflow != nil {
  workflow?.append(item.metadata)
} else {
  workflow = AnyWorkflow(Workflow<FR>(item.metadata))
}
```

But after looking at that and not being able to make it an `if let workflow = workflow`, I think I'm coming around that `if workflow == nil` is clearer.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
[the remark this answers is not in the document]
Richard-Gist: Made it private.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
Richard-Gist: Removed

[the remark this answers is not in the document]
Richard-Gist: Removed

Sources/SwiftCurrent_SwiftUI/WorkflowItem.swift
Richard-Gist: made private

SwiftCurrent.podspec
Tyler-Keith-Thompson: Take a look at the commit/action history. I started with 14, then CocoaPods cried cause the subspec had a higher version than the primary spec. I suspected there was no way around it, so downgraded to 11 to make the primary spec

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+        isPresented?.wrappedValue = false
+        onAbandon.forEach { $0() }
+    }
+
+    func complete(_ workflow: AnyWorkflow, passedArgs: AnyWorkflow.PassedArgs, onFinish: ((AnyWorkflow.PassedArgs) -> Void)?) {
+        if workflow.lastLoadedItem?.value.metadata.persistence == .removedAfterProceeding {
+            if let lastPresentableItem = workflow.lastPresentableItem {
+                extractView(from: lastPresentableItem).model = self
+            } else {
+                #warning("We are a little worried about animation here")
```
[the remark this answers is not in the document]
Tyler-Keith-Thompson: Suspect we can resolve it by setting `isPresented` to false. I was originally against it but in retrospect I think it makes sense here.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
```diff
@@ -0,0 +1,214 @@
...
+                didLoad = true
+                model.isPresented = $isPresented
+                model.onAbandon = onAbandon
+                workflow?.launch(withOrchestrationResponder: model,
+                                 passedArgs: passedArgs,
+                                 launchStyle: .new) { passedArgs in
+                    onFinish.forEach { $0(passedArgs) }
+                }
+            }
+            .onDisappear {
```
Richard-Gist, line 101: I think we need to reset the didLoad here back to false.  The reason is that if you toggle showing the workflow, we end up failing to show it when we toggle back on.
```suggestion
            .onDisappear {
                didLoad = false
```

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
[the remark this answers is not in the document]
Richard-Gist: Yes this is a problem (confirmed) and probably for the `isPresented`.

wiemerm:
EOD: There are still some gaps with how `onDisappear` works in `WorkflowView` that we are working through. Some examples include relaunching a workflow after the `isPresented` binding Bool is toggled and handling when scrolled away within a `ScrollView` where `onDisappear` gets called.

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
```diff
@@ -0,0 +1,68 @@
...
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//
+
+import SwiftCurrent
+import SwiftUI
+
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+final class WorkflowViewModel: ObservableObject {
+    @Published var body = AnyView(EmptyView())
+    var isPresented: Binding<Bool>?
```
Tyler-Keith-Thompson, line 15: I think `isPresented` is not a very accurate description of what this binding does. Case in point if you want to relaunch your workflow you toggle it off, then on. 

I don't necessarily have a great suggestion for a replacement but I'll humbly suggest `isLaunched` in place of `isPresented`

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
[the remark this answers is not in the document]
Richard-Gist: Switched to the isPresented but we should discuss what it means to flip that Binding.

Tests/SwiftCurrent_SwiftUITests/PersistenceTests.swift
Richard-Gist: This isn't really a problem in our tests at this time, so I think we'll leave it for now.

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
Richard-Gist: Here are some iterations:
```swift
WorkflowView(isPresented: .constant(true)) // Current
WorkflowView(isLaunched: .constant(true)) // Suggested by Tyler
WorkflowView(launch: .constant(true))
WorkflowView(run: .constant(true))
WorkflowView(activate: .constant(true))
```

Richard-Gist: I think i'm leaning towards `isLaunched`

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
```diff
@@ -0,0 +1,215 @@
...
+    private var onFinish = [(AnyWorkflow.PassedArgs) -> Void]()
+    private var onAbandon = [() -> Void]()
+    private var passedArgs = AnyWorkflow.PassedArgs.none
+
+    /// Creates a `WorkflowView` that displays a `FlowRepresentable` when presented.
+    public init(isPresented: Binding<Bool>) where Args == Never {
+        _isPresented = isPresented
+    }
+
+    /// Creates a `WorkflowView` that displays a `FlowRepresentable` when presented.
```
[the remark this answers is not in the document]
Richard-Gist:
```swift
startingArgs
startingArguments
```

Tyler-Keith-Thompson:
@Richard-Gist Your comment wasn't a code review so I can't reply to it:
> With this extension, do we need to put anything in UIKit or Core to test this?

No I don't think so, I mean we *can* but we have a test for it, so I'm satisfied. Any additional tests would be redundant. I much prefer the black-box style testing we've got because it allows for maximum refactoring but still gives confidence.

Tests/SwiftCurrent_SwiftUITests/PersistenceTests.swift
```diff
@@ -0,0 +1,521 @@
...
+        }
+        let expectOnFinish = expectation(description: "OnFinish called")
+        let expectViewLoaded = ViewHosting.loadView(
+            WorkflowView(isPresented: .constant(true))
+                .thenProceed(with: WorkflowItem(FR1.self).persistence(.persistWhenSkipped))
+                .thenProceed(with: WorkflowItem(FR2.self).persistence(.persistWhenSkipped))
+                .thenProceed(with: WorkflowItem(FR3.self).persistence(.persistWhenSkipped))
+                .thenProceed(with: WorkflowItem(FR4.self).persistence(.persistWhenSkipped))
+                .onFinish { _ in expectOnFinish.fulfill() })
+            .inspection.inspect { _ in }
```
brianlombardo, line 517: Will add `assert` to validate that `FR4` is visible

Tests/SwiftCurrent_SwiftUITests/WorkflowViewModelTests.swift
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+}
+
+fileprivate struct FR: FlowRepresentable {
+    var _workflowPointer: AnyFlowRepresentable?
+}
+
+extension FlowRepresentableMetadata {
+    static func createForTests<FR: FlowRepresentable>(_: FR.Type) -> FlowRepresentableMetadata {
```
brianlombardo, line 54:
```suggestion
    private static func createForTests<FR: FlowRepresentable>(_: FR.Type) -> FlowRepresentableMetadata {
```

Tests/SwiftCurrent_SwiftUITests/WorkflowViewModelTests.swift
```diff
@@ -0,0 +1,64 @@
...
+        .init(FR.self, flowPersistence: { _ in .default })
+    }
+}
+
+extension AnyWorkflow.Element {
+    static func createForTests<FR: FlowRepresentable>(_ :FR.Type) -> AnyWorkflow.Element {
```
brianlombardo, line 60:
```suggestion
    private static func createForTests<FR: FlowRepresentable>(_ :FR.Type) -> AnyWorkflow.Element {
```

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -0,0 +1,479 @@
...
+            WorkflowView(isPresented: .constant(true), args: AnyWorkflow.PassedArgs.args(expected))
+                .thenProceed(with: WorkflowItem(FR1.self))
+                .thenProceed(with: WorkflowItem(FR1.self))).inspection.inspect { viewUnderTest in
+            XCTAssertEqual(try viewUnderTest.find(FR1.self).actualView().property.extractArgs(defaultValue: nil) as? String, expected)
+        }
+
+        wait(for: [expectViewLoaded], timeout: 0.3)
+    }
+
+    func testWorkflowPassesArgumentsToTheAllItems() throws {
```
brianlombardo, line 129:
```suggestion
    func testWorkflowPassesArgumentsToAllItems() throws {
```

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -0,0 +1,479 @@
...
+        let binding = Binding(wrappedValue: true)
+        let workflowView = WorkflowView(isPresented: binding)
+            .thenProceed(with: WorkflowItem(FR1.self)
+                            .applyModifiers { $0.customModifier().background(Color.blue) })
+        let expectViewLoaded = ViewHosting.loadView(workflowView).inspection.inspect { viewUnderTest in
+            // Capture vstack for lifecycle simulation
+            let vstack = try viewUnderTest.vStack()
+            binding.wrappedValue = false
+            XCTAssertNoThrow(try vstack.callOnDisappear())
+            // Expected view at this point in the lifecycle
```
brianlombardo, line 350:
```suggestion
            // Expected that there is no vstack at this point in the lifecycle
```

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -0,0 +1,479 @@
...
+        }).inspection.inspect { viewUnderTest in
+            XCTAssertNoThrow(try viewUnderTest.find(FR1.self).actualView().proceedInWorkflow())
+            XCTAssertNoThrow(try viewUnderTest.vStack().callOnAppear())
+            XCTAssertNoThrow(try viewUnderTest.find(FR2.self).actualView().proceedInWorkflow())
+        }
+
+        wait(for: [expectOnFinish, expectViewLoaded], timeout: 0.3)
+    }
+    
+    func testWorkflowRelaunchesWhenSubsequentlyPresented() throws {
```
brianlombardo, line 387:
```suggestion
    func testWorkflowRelaunchesWhenSubsequentlyLaunched() throws {
```

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -0,0 +1,479 @@
...
+                    binding.wrappedValue = true
+                    XCTAssertNoThrow(try viewUnderTest.vStack().callOnAppear())
+                    XCTAssertNoThrow(try viewUnderTest.find(FR1.self).actualView().proceedInWorkflow())
+                    XCTAssertNoThrow(try viewUnderTest.find(FR2.self).actualView().proceedInWorkflow())
+                }
+        
+        wait(for: [expectOnFinish, expectViewLoaded], timeout: 0.3)
+    }
+
+    func testWorkflowMaintainsStateWhenViewDisappearsAndReappears_WithoutIsPresentedChanging() throws {
```
brianlombardo, line 418:
```suggestion
    func testWorkflowMaintainsStateWhenViewDisappearsAndReappears_WithoutIsLaunchedChanging() throws {
```

brianlombardo (changes requested):

morganzellers (dismissed):

Sources/SwiftCurrent_SwiftUI/Views/WorkflowView.swift
```diff
@@ -0,0 +1,236 @@
...
+    @Binding private var isPresented: Bool
+    @StateObject private var model = WorkflowViewModel()
+    @State private var didLoad = false
+
+    let inspection = Inspection<Self>() // Needed for ViewInspector
+    private var workflow: AnyWorkflow?
+    private var onFinish = [(AnyWorkflow.PassedArgs) -> Void]()
+    private var onAbandon = [() -> Void]()
+    private var passedArgs = AnyWorkflow.PassedArgs.none
+
```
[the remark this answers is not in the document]
Richard-Gist: Sadly no, not without further refactoring around initialization.

wiemerm (approved):

Richard-Gist (approved):

morganzellers (approved):

brianlombardo (approved):

PR #73, opened 2021-07-15
Swiftui example app

Tyler-Keith-Thompson:
## Checklist:

----

### If Applicable:

----


Richard-Gist:
EOD: No real progress has been made on this.  Debugging the failures that only happen in the pipeline is proving difficult.

Richard-Gist:
EOD: I think this is it.

If it all looks good, then let's get it merged into trunk.  If there are tweaks we want to make, then we may want to also add some linking in README to the other README's.  Those are optimizations we could make in another PR though.

ExampleApps/SwiftUIExample/SwiftUIExampleTests/TestConstant.swift
```diff
@@ -0,0 +1,16 @@
...
+//  Created by Richard Gist on 7/20/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//
+
+import Foundation
+
+enum TestConstant {
+    static var timeout: TimeInterval {
+        10.0 // Pipeline
+//        return 0.1; #warning("Local timeout should not be checked in")
```
wiemerm, line 14: 10s timeout?

README.md
```diff
@@ -30,15 +30,15 @@ As you continue to develop your applications, each view controller will become m
 
 ## See it in action with our example app
 
-Clone our repo, open `SwiftCurrent.xcworkspace`, target the `SwiftCurrentExample` scheme, and run to see our example app in action.
+Clone our repo, open `SwiftCurrent.xcworkspace`, target the `UIKitExample` scheme, and run to see our example app in action.
 
-The app is designed to give you an idea of what SwiftCurrent can do with minimal overhead in the UI.  The key areas of code you'll want to look at are: `SetupViewController.swift` and the view controllers referenced in the [Workflow](https://wwt.github.io/SwiftCurrent/Classes/Workflow.html).
+The app is designed to give you an idea of what SwiftCurrent can do with minimal overhead in the UI. [See the README](https://github.com/wwt/SwiftCurrent/blob/8e11ce58cc5410d09a40d43fefc0e722c87906b3/ExampleApps/UIKitExample/README.md) for the example project for suggestions of where to start looking in the project.
```
wiemerm, line 35: This second sentence reads awkwardly. Not even sure I get what the message is that it's trying to say.

Should I be looking for a README in the example project for suggestions of where to start looking in the project?
Is it look at the README and the example project?
Or am I looking for references to the example project in the README?

wiemerm (changes requested):
Several test files switch the wait from 0.3 to 0.5. Was this truly necessary? I'm ok with it at this point if it was.

ExampleApps/SwiftUIExample/SwiftUIExampleTests/TestConstant.swift
Richard-Gist: I didn't like that either.  The good news is that this 10 second timeout only applies in this context when ViewInspector is unable to call `.inspect` and the places where we are testing for `onFinish` or `onAbandon` to be called.  Both of these scenarios will only hit the 10 seconds if they completely fail.

Richard-Gist: In my investigation of the pipeline, we could prune it down to 5 seconds and still be pretty safe.  3 seconds would be mostly safe, as I didn't see any test run all the way up to 3 seconds, but 1 test did get close.

Richard-Gist: A heavier lift option would be to split the waiting for expectations into 2 lines.  There would be a long timeout wait **just** for the `inspect` call, and a shorter timeout on the `onFinish`.  If we wait for `inspect` then wait for `onFinish` then you would have up to 10 seconds to wait on inspecting (which should always pass) and then a shorter .5 second timeout that onFinish was called (which will probably already be called at this point).

Drawbacks to this plan: If you do the order wrong the test will become flakey again, Quite a bit of code change and will probably be forgotten in future tests. The longest wait will be 10.5 seconds for if everything catastrophically fails.

Positives: Passing tests will be unnoticeably longer, Failing tests for when something breaks in `onLoad` or `onAbandon` will happen quickly (expected to be .5 + time it took to inspect).

README.md
Richard-Gist: Yeah I'm not fully sure what I was trying to say either.  We now have README files in each example project that gives an overview of what the project is and interesting areas to start looking at.  For example, SwiftUI calls out the 3 main features, and then calls out ContentView as the entry point launching the workflows and being a tab item example.

So I'm trying to link to that README so people on Pages and people visiting Github know that there is a README.  Those who have opened the workspace and are looking at the Example project will see the README at the root level, inside the project.

Richard-Gist: Unrelated to this issue, the 0.3 second change apparently was not really needed.  It was to make the flakey Model tests to pass more reliably, I believe.  We could change it back, though our next item to work is fixing those flakey tests.  Your call.

ExampleApps/SwiftUIExample/SwiftUIExampleTests/TestConstant.swift
wiemerm: I'd like to at least try to get it down if possible considering there appear to be quite a few places that I saw using that `timeout`. If there's nothing we can really do about it other than wait out the inspect call then it is what it is, and I can accept that.

ExampleApps/SwiftUIExample/SwiftUIExampleTests/TestConstant.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Richard Gist on 7/20/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//
+
+import Foundation
+
+enum TestConstant {
+    static var timeout: TimeInterval {
+        10.0 // Pipeline
```
morganzellers, line 13:
```suggestion
        5.0 // Pipeline
```

ExampleApps/SwiftUIExample/SwiftUIExampleTests/TestConstant.swift
Richard-Gist: We can prune it down to 5 seconds at least.  @morganzellers has already suggested that change in his review so I'll take that.

README.md
Richard-Gist: How about:
```suggestion
The [example app has a README](https://github.com/wwt/SwiftCurrent/blob/8e11ce58cc5410d09a40d43fefc0e722c87906b3/ExampleApps/UIKitExample/README.md) that details interesting usages.
```

morganzellers (approved):

wiemerm (approved):

PR #77, opened 2021-07-19
[Beta] Removes usage of AnyView

Tyler-Keith-Thompson:
## Checklist:

----

### If Applicable:

----


Sources/SwiftCurrent_SwiftUI/Views/ModifiedWorkflowView.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    public var body: some View {
+        if isLaunched {
+            if let body = model.body as? Content {
+                body
+                    .onReceive(model.onAbandonPublisher) { onAbandon.forEach { $0() } }
+                    .onChange(of: isLaunched) { if $0 { launch() } }
+                    .onReceive(inspection.notice) { inspection.visit(self, $0) }
+            } else {
+                wrapped.onReceive(inspection.notice) { inspection.visit(self, $0) }
```
Richard-Gist, line 35: I'm not sure this works but if it does I'd like:
```suggestion
                wrapped?.onReceive(inspection.notice) { inspection.visit(self, $0) }
```
To make it clear that that is optional.

Sources/SwiftCurrent_SwiftUI/WorkflowItem.swift
```diff
@@ -30,32 +30,42 @@ import SwiftCurrent
...
+        modifierClosure = modifier
+    }
+
     /**
      Provides a way to apply modifiers to your `FlowRepresentable` view.
 
      ### Important: The most recently defined (or last) use of this, is the only one that applies modifiers, unlike onAbandon or onFinish.
      */
-    public func applyModifiers<V: View>(@ViewBuilder _ closure: @escaping (F) -> V) -> Self {
+    public func applyModifiers<V: View>(@ViewBuilder _ closure: @escaping (F) -> V) -> WorkflowItem<F, V> {
```
Richard-Gist, line 59: This is no longer `Self`?  But it looks like `Self`.

SwiftCurrentLint/.swiftlint.yml
```diff
@@ -132,7 +132,7 @@ excluded: # paths to ignore during linting. Takes precedence over `included`.
   - Workflow/Package.swift
 
 line_length:
-  warning: 191
+  warning: 196
```
Richard-Gist, line 135: @wiemerm You good on this?  I'm ok with it.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -228,6 +238,36 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
         wait(for: [expectViewLoaded], timeout: 0.3)
     }
 
+    func testWorkflowOnlyShowsOneViewAtATime() throws {
```
Richard-Gist, line 241: I'm not sure this test name matches the asserts.  If you showed all the views at the same time, wouldn't the test down below still pass?

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -324,154 +364,118 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
...
-                    binding.wrappedValue = true
-                    XCTAssertNoThrow(try viewUnderTest.vStack().callOnAppear())
-                    XCTAssertNoThrow(try viewUnderTest.find(FR1.self).actualView().proceedInWorkflow())
-                    XCTAssertNoThrow(try viewUnderTest.find(FR2.self).actualView().proceedInWorkflow())
-                }
-        
-        wait(for: [expectOnFinish, expectViewLoaded], timeout: 0.3)
-    }
 
-    func testWorkflowMaintainsStateWhenViewDisappearsAndReappears_WithoutIsLaunchedChanging() throws {
```
Richard-Gist, line 418: Git might be making it hard to track where this test went, but this is still a valid-ish test (probably not the way it was being tested).  It's important that if our views roll off the screen and then come back, they haven't reset their state.

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
```diff
@@ -8,37 +8,43 @@
...
 
     func backUp(from source: AnyWorkflow.Element, to destination: AnyWorkflow.Element) {
-        extractView(from: destination).model = self
+        body = extractView(from: destination).erasedView
     }
 
     func abandon(_ workflow: AnyWorkflow, onFinish: (() -> Void)?) {
         isLaunched?.wrappedValue = false
-        onAbandon.forEach { $0() }
+        body = nil
```
Richard-Gist, line 40: I'm not sure I understand this part.  Why `nil`?

Sources/SwiftCurrent_SwiftUI/Views/ModifiedWorkflowView.swift
```diff
@@ -0,0 +1,165 @@
...
+//  Created by Tyler Thompson on 7/20/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//
+//  swiftlint:disable file_types_order
+
+import SwiftUI
+import SwiftCurrent
+
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+public struct ModifiedWorkflowView<Args, Wrapped: View, Content: View>: View {
```
Richard-Gist, line 14: I missed it.  I'm not sure how this ended up on the screen.  Where did this body get put on the view tree?

Richard-Gist (commented):
How much testing is left on this?

Sources/SwiftCurrent_SwiftUI/WorkflowItem.swift
Tyler-Keith-Thompson: Looks like, but isn't The trick here is that `V` representing the result of the modifier block is now stored on the `WorkflowItem`. Previously the second specialization of `WorkflowItem` was almost certainly the `FlowRepresentable & View` but now it it's most likely a `ModifiedContent<..., FlowRepresentable & View>`

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: Nope, the trick is at the bottom. ViewInspector doesn't support `find(FR1.self, skipFound 1)` so I used `ViewType.Text.self`. This is in relation to #76 where I talked about how if you don't have an `if` statement for casting the body when you have the same `FlowRepresentable` twice in a `Workflow` it presents both of them. This test proves that does *not* happen.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: Correct, and that's a function of SwiftUI at this point. Because we do not have an `onAppear` or `onDisappear` modifier that change anything all behavior will be as Apple has intended. BOW DOWN TO THEIR IDEAS

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
Tyler-Keith-Thompson: Because the `Workflow` is now gone, so any remnants of the views should be gone as well.

Tyler-Keith-Thompson: I'll leave this one unresolved in case that answer is somehow unsatisfactory...Suspect this would be pretty easy to understand with animation cruft.

Sources/SwiftCurrent_SwiftUI/Views/ModifiedWorkflowView.swift
Tyler-Keith-Thompson: `thenProceed(with:` from a `WorkflowView` creates on of these bad boys.

Tyler-Keith-Thompson:
EOD: We have obliterated AnyView which is just a beautiful feeling. At this point we believe this is ready to merge to trunk but would recommend merging the example app first. That'll give a test bed to make sure things still work and give us the ability to play around with animations n' such. 

AnyView has been shot and buried out back.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Richard-Gist: I'll bow down to their ideas.  In doing so, I'd like to confirm that if WorkflowView is placed in a lazyVStack and scrolled off the screen and then comes back into view, the user is still in the same spot of their workflow.  A test would be amazing, but also just confirming that has been tested would be enough in this situation.

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
Richard-Gist: So a `nil` view is different than an `EmptyView`?

Sources/SwiftCurrent_SwiftUI/Views/ModifiedWorkflowView.swift
Richard-Gist: That instantiates a `ModifiedWorkflowView`.  I meant my question more towards, I don't see how the body of a `ModifiedWorkflowView` makes it up to the `ContentView` that created the `WorkflowView`.  `WorkflowView` always has a `fatalError` in its body, so I'm missing the connection between `ContentView { var body: some View }` and `ModifiedWorkflowView` body.

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
Tyler-Keith-Thompson: Absolutely `nil` is the absence of a value. With our recursive crazy this essentially means nothing is provided as a View to display, allowing for maximum SwiftUI jiggery pokery optimization from Apple. `EmptyView` is a concrete type meaning "I don't have a value". I suspect, but do not know for certain, that it probably itself is just an `if false` statement.

Richard-Gist:
🧟

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: I'm unaware of how to do such a test without pulling in XCUITest and a hosting application. If that flakified test is worthwhile to you we can do it, elsewise it might be just a matter of pulling in the example app and proving it there.

Sources/SwiftCurrent_SwiftUI/Views/ModifiedWorkflowView.swift
Richard-Gist: Wait ok, I think it's clicking.  The `thenProceed(with:` returns a `ModifiedWorkflowView`, thus replacing my `WorkflowView` with a `ModifiedWorkflowView` and 💥  the `WorkflowView` is not on the view graph.

Tyler-Keith-Thompson: Winner winner chicken dinner. That's precisely it.

Sources/SwiftCurrent_SwiftUI/Models/WorkflowViewModel.swift
Richard-Gist: 👍  I wonder what animation weirdness this is going to cause, but I wonder that for everything SwiftUI.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Richard-Gist: Ok.  Let's leave this open until someone validates it.  I think we'll be good enough with just a solid check of this (100 items or more would be preferable).  No flakey test needed.

Whoever validates this please respond here and resolve the conversation.

Tyler-Keith-Thompson: Confirmed that this works exactly as expected with a LazyVStack and looooots of scrolling.

Richard-Gist (changes requested):
Hinky-ness in Profile MFAView.  Seems like there's some conflicting ID stuff, I think.

SwiftCurrentLint/.swiftlint.yml
wiemerm: Yes, I'm good with this.

Tyler-Keith-Thompson:
Update: It would appear that it's not necessarily ID related, that maybe there's just a bug when showing 2 workflow views at all. I can't help but wonder if it has something to do with a model getting shared that should not be? Like maybe *everybody* is reading from the same erasedBody?

To prove it out: Just remove MFAView and run, you'll see that even though one is a ChangePasswordView and one is a ChangeUsernameView they cannot both display at the same time. You can also add all the `id`s you want and it'll never change. 

Not entirely sure what that bug is, but I think it might be easier than tracking down an ID thing.

Richard-Gist:
EOD: Launcher seems to be the problem.  See the debug branch: anyview-debug

Tyler-Keith-Thompson:
My EOD: So I played with this a bunch and got a working solution. The `WorkflowView` behaves as you'd expect showing both workflows on the Profile screen without any runtime warnings or errors. 

The SwiftCurrent_SwiftUI tests all pass but lie badly about StateObject being misused. DO NOT LISTEN TO THEM. This is actually a ViewInspector thing see here: https://github.com/nalexn/ViewInspector/blob/master/readiness.md#property-wrappers StateObject isn't inspectable. Don't take my word for it though, run the example app and witness the lack of issues, then run the tests and look at the trace for the `StateObject` runtime error, it's entirely caused by ViewInspector.

The SwiftUIExample tests are failing on the multi-workflow tests. This seems to have something to do with onFinish and a state change? I'm really not quite certain at this point. Breakpoints on the view suggest that it's fine and that the tests are lying. 


An update on what I did:
- Turns out changing it to use `@StateObject` now was really easy...I literally just replaced `@ObservedObject` with `@StateObject` and it worked. Entirely due to the changes we made after the first time we attempted this.
- Changing it to use `@StateObject` ALSO solved all the problems with multiple workflows on the same screen (ala: Profile). However it caused one additional problem, a production-code runtime error for anything with an `onFinish` modifier
- I tweaked how `onFinish` works so that it's publisher based, much like how `onAbandon` works. Which honestly felt good anyways.
- I then started running into test issues where `onFinish` was getting called twice for some views and not for others. The short version is it had to do with `wrapped?.onReceive(onFinish) // <--- NECESSARY` and `body.onReceive(onFinish) // <--- ALSO NECESSARY`
- To fix this I simply used our `launcher` `@StateObject` and added a boolean to it indicating whether we'd already called `onFinish` or not, easy peasy
- For workflows where literally every item is skipped `onFinish` gets called before the view renders and the `onReceive` block can trigger, meaning it was never triggering. To fix that I simply changed the kind of publisher from a `PassthroughSubject` to a `CurrentValueSubject` which, for the record stores it's values and sends them to subscribers when they subscribe
- Because `CurrentValueSubject` *requires* a value at all times I had to initialize it with *something* for our onFinish call, so I initialized it with `nil` (as opposed to AnyWorkflow.PassedArgs.none` because that let my `onFinish` subscribers know whether it was really finished or whether it just received the first ever value (nil)

Tyler-Keith-Thompson:
ALSO NOTE: Look at the coverage report. I think one of the tests we deleted a while back because we didn't need it was covering one of our `thenProceed(with:` calls on `WorkflowView`. We should make sure and add a new test that expressly covers that use case. And of course, any other gaps in coverage.

Tyler-Keith-Thompson:
> Have you done everything you can to make sure that errors that can occur are compile-time errors, and if they have to be runtime do you have adequate tests and documentation around those runtime errors? For more details.

This question is "no" with the current implementation. We can easily make this happen though and have already determined the consumer experience is much better. 

I'm holding off on commit while we come up with a name. Proposals:
Change WorkflowView to WorkflowLauncher (it's not a view anymore so that compiler errors happen if nobody calls `thenProceed(with:)`

... that's it so far.

Tyler-Keith-Thompson:
You cannot have a PR about [recursion](https://github.com/wwt/SwiftCurrent/pull/77) without

Richard-Gist (approved):
Looks good enough to Demo.  Anything missed, we could open an issue on and solve it.

brianlombardo (approved):

PR #79, opened 2021-07-24
Adds Passthrough representable

### LINKED ISSUE #21: `Passthrough` FlowRepresentable
Not so much a problem, just an idea

Idea: We could have a Passthrough type that's a little different than Never. Whereas Never means "I don't take in a value" or "I don't pass out a value". Passthrough could mean "I'll forward on anything that gets passed to me automatically`.

Hypothetical: Imagine a workflow wherein you've got a terms and conditions FlowRepresentable. Now it knows whether terms have been accepted or not, and it will make the decision to abandon whatever workflow it is in if the terms are not accepted. However if terms are accepted it might send that information to a server then move forward.

If that termsAndConditions had a Passthrough type for its input and output it would be even easier to stick those terms and conditions in any workflow while maintaining type safety of that flow.

I mean users can do this now, manually. So they could continue to do it manually. Something like:
```swift
struct FR1: FlowRepresentable {
    typealias WorkflowOutput = AnyWorkflow.PassedArgs
    var _workflowPointer: AnyFlowRepresentable?
    var passedArgs: AnyWorkflow.PassedArgs
    init(with args: AnyWorkflow.PassedArgs) {
        passedArgs = args
    }
    
    func doAThing() {
        proceedInWorkflow(passedArgs)
    }
}
```

That's just a lot of boilerplate

Tyler-Keith-Thompson:
## Checklist:
- [x] Do we need to [increment the minor/major version](https://github.com/wwt/SwiftCurrent/blob/main/.github/CONTRIBUTING.md#versioning)?

----

- [x] Did you deprecate (rather than remove) any old methods/variables/etc? [Our philosophy for deprecation](https://github.com/wwt/SwiftCurrent/blob/main/.github/CONTRIBUTING.md#deprecation).


Tyler-Keith-Thompson:
### Alternative Approaches Considered:
- There was an approach where a new type could be declared called `Passthrough`. This idea was discarded for 2 important reasons, the first is that it'd only work if both `Workflowinput == Passthrough` and `WorkflowOutput == Passthrough`, but there was no reasonable way to enforce that always happened. The second is that the generic constraints we have in place are already sort of hard to wrap your head around, they would've positively exploded with a new type like that.
- There was the possibility of making an input and output of `Never` mean `Passthrough` but this was rejected because `Never` has a specific connotation right now (if my input type is `Never` it means I'll happily drop your arguments on the floor). This is purposeful because it may be a legitimate usecase in a `Workflow`. So abusing `Never` to somehow work with this felt wrong.
- There was the possibility of looking at both the `Input` and `Output` of a `FlowRepresentable` and if they were both `AnyWorkflow.PassedArgs` considering that to mean `Passthrough` but that was discarded because it felt like the behavior of automatically passing data through that `FlowRepresentable` was a bit magic.

### Design Choices:
- A new protocol was created named `PassthroughFlowRepresentable`, while it is a lengthy name consumers of the library would be very clear in their intent and would be aware of what that does (mostly, remove boilerplate)
- Very special and deliberate attention was payed to the current tests and current generic constraints. They were not fully fleshed out, so a big part of this PR was adding tests around how things should've behaved already. To that end `GenericConstraintTests` have been added to `SwiftCurrent`, `SwiftCurrent_UIKit`, and `SwiftCurrent_SwiftUI`
- Because `thenProceed(with:)` reads so much better than `thenPresent` and because `thenPresent` means a whole different series of generic constraints it was decided to deprecate (WARNING, not error) those methods and use the `rename` attribute to let Xcode tell consumers how to update
- There's a commit in here called "Xcode driven development" where after that deprecation was added I just clicked "fix" on Xcode and it correctly applied all changes, which was beautiful
- For SwiftUI ONLY* Because our SwiftUI implementation differentiates the first item in the workflow from the rest there was a deliberate choice made not to support arguments in a `WorkflowLauncher` and the very next `thenProceed(with:)` having an input type of `Never`. If users *really* want the behavior of launching a workflow with arguments and then IMMEDIATELY dropping them on the floor they can make their first `FlowRepresentable` take a `WorkflowInput` of `AnyWorkflow.PassedArgs` and simply not pass them forward. However it seemed there was no legitimate reason to allow this, it feels very much like an error case.
- For SUBSEQUENT calls to `thenProceed(with:)` if one item passes forward data and the next has a `WorkflowInput` of `Never` the old behavior is honored and that data is dropped on the floor. There may be several legitimate reasons for doing this, there may be skips that happen, there may be a reason to transform that data. It makes a lot more sense once you've gotten past the first item in a `Workflow`
- Generic constraints have been pushed probably to their limit with UIKit. It seems more clear than ever that we should try an approach where we use a modifier-like syntax for `launchStyle` and `persistence`. However, that felt WAY outside of the scope of this PR. Once that decision is made you should be able to deprecate some of the methods on `Workflow` now and rename them much like `thenPresent` was deprecated as part of this PR

Tyler-Keith-Thompson:
### NOTE:
- Look at how cool `MFAView` is now, WAY less boilerplate, same experience

Tyler-Keith-Thompson:
### Rules for Generic Constraints

- If a `WorkflowInput` is `Never` it means either no arguments or some arguments can be passed to the `FlowRepresentable`, but any arguments passed will be ignored and will not be passed to the next item.
- if a `WorkflowInput` is `AnyWorkflow.PassedArgs` it means either no arguments or some arguments can be passed to the `FlowRepresentable`, and it'll handle those arguments somehow, either by passing them on, transforming them, or dropping them
- If a `WorkflowInput` is a concrete type (like say, a `String`) it means a `FlowRepresentable` will only take that type, if something that isn't a `String` is passed to it, or if nothing is passed to it, there'll be a runtime error.
` if a `FlowRepresentable` is a `PassthroughFlowRepresentable` it means it will not deliberately handle arguments, it's `WorkflowInput` and `WorkflowOutput` are *both* considered to be an `AnyWorkflow.PassedArgs` but the library will handle passing its arguments forward. These `PassthroughFlowRepresentable`s simply call `proceedInWorkflow()` and do not specify arguments.

### In UIKit

#### Valid Workflows FORMAT: {LAUNCH_ARGS}  >> INPUT>Output ~> Next Item
- {String}>> String>Never ~> AnyWorkflow.PassedArgs>AnyWorkflow.PassedArgs ~> Never>Never
- AnyWorkflow.PassedArgs ~> String>Never (NOTE: Fails if argument passed was not a string, or was `.none`)
- {AnyWorkflow.PassedArgs}>> String>String ~> String>Int ~> Int>Never
- {AnyWorkflow.PassedArgs}>> String>String ~> PASSTHROUGH ~> String>Int

#### Compile time works, but fails at runtime:
- {Int}>> String>String ~> String>Int ~> Int>Never
- {Int}>> Int>String(SKIPPED) ~> String>Int ~> Int>Never
- {Never}>> String>String ~> String>Int ~> Int>Never

### In SwiftUI

#### Valid Workflows FORMAT: {LAUNCH_ARGS}  >> INPUT>Output ~> Next Item
- {String}>> String>Never ~> AnyWorkflow.PassedArgs>AnyWorkflow.PassedArgs ~> Never>Never
- AnyWorkflow.PassedArgs ~> String>Never (NOTE: Fails if argument passed was not a string, or was `.none`)
- {AnyWorkflow.PassedArgs}>> String>String ~> String>Int ~> Int>Never
- {AnyWorkflow.PassedArgs}>> String>String ~> PASSTHROUGH ~> String>Int
- {Never}>> Never>String ~> Never>Never

#### Compile time works, but fails at runtime:
- {Int}>> Int>String(SKIPPED) ~> String>Int ~> Int>Never

#### Invalid Workflows that could've worked in UIKit
- {Never}>> String>String ~> String>Int ~> Int>Never

Richard-Gist (commented):
Looks good so far, but I want to play with it a bit before I sign off.

Richard-Gist:
The failed test after the merge was due to an exceeded timeout in `SwiftCurrent_SwiftUITests.PersistenceTests testRemovedAfterProceeding_OnAllItemsInAWorkflow` which exceeded the timeout of 0.5 seconds for the `Inspection at line 189`  The total test time for the failed test was `4.886 seconds`.  The logs are not showing me anything more of interest, other than all tests ran slow (but not that slow).  Re-running the tests to pass as I think this was a system hiccup.

Tyler-Keith-Thompson:
Missing: PassthroughFlowRepresentable does not yet play nice with StoryboardLoadable

Richard-Gist (changes requested):
Having PassthroughFlowRepresentable work with StoryboardLoadable is going to be a requirement for me.  Without it, I can't use it in our example app to showcase a simple way to leverage it.

Richard-Gist:
So I'm trying to test it with our example app:
```swift
class TermsOfServiceViewController: UIViewController, PassthroughFlowRepresentable, StoryboardLoadable {
    weak var _workflowPointer: AnyFlowRepresentable?

    @IBAction private func accept(_ sender: Any) {
        proceedInWorkflow()
    }
}
```

And when I add it to the SetupViewController:
```swift
launchInto(
    Workflow(LocationsViewController.self)
        .thenProceed(with: PickupOrDeliveryViewController.self)
        .thenProceed(with: TermsOfServiceViewController.self) // <----- just added it in the middle somewhere
        .thenProceed(with: MenuSelectionViewController.self, flowPersistence: .persistWhenSkipped)
        .thenProceed(with: FoodSelectionViewController.self)
        .thenProceed(with: ReviewOrderViewController.self),
    args: locations,
    withLaunchStyle: .navigationStack)
```

I get a fatalError when it is time to show the screen:
`SwiftCurrent_UIKit/StoryboardLoadable.swift:78: Fatal error: The StoryboardLoadable protocol provided a default implementation if this initializer so that consumers didn't have to worry about it in their UIViewController. If you encounter this error and need this initializer, simply add it to TermsOfServiceViewController`

You may find it helpful to go ahead and add a ToS screen to the example app to test out consuming it in a Storyboards example.  It would also give your feature a cool shoutout on why it's awesome.

Sources/SwiftCurrent_UIKit/StoryboardLoadable.swift
```diff
@@ -70,6 +70,18 @@ extension StoryboardLoadable {
     }
 }
 
+@available(iOS 13.0, *)
+extension PassthroughFlowRepresentable where Self: StoryboardLoadable {
+    /// :nodoc: **WARNING: This will throw a fatal error.** Just a default implementation of the required `FlowRepresentable` initializer meant to satisfy the protocol requirements.
+    public init(with args: WorkflowInput) { // swiftlint:disable:this unavailable_function
+        // swiftlint:disable:next line_length
+        fatalError("The StoryboardLoadable protocol provided a default implementation if this initializer so that consumers didn't have to worry about it in their UIViewController. If you encounter this error and need this initializer, simply add it to \(String(describing: Self.self))")
```
Richard-Gist, line 78:
```suggestion
        fatalError("The StoryboardLoadable protocol provided a default implementation of this initializer so that consumers didn't have to worry about it in their UIViewController. If you encounter this error and need this initializer, simply add it to \(String(describing: Self.self))")
```

Richard-Gist (commented):
Found a typo.  If this was a copy paste thing, it's ok to just update this code, but updating where you copied it from would be appreciated.

Sources/SwiftCurrent_UIKit/StoryboardLoadable.swift
Tyler-Keith-Thompson: No need, this whole init is gone now

Richard-Gist:
I'm still having trouble using this in the UIKitExample app.  Using the code I posted above, here's the results:

https://user-images.githubusercontent.com/79471462/127044741-b5105bd2-71e5-42c8-9d28-09745edae543.mp4

Tyler-Keith-Thompson:
Woah...there was a deep underlying issue with SwiftCurrent altogether here. Short version? When the input type was `AnyWorkflow.PassedArgs` it was not using the factory method. Meaning all the `StoryboardLoadable` protocol things simply don't work with `AnyWorkflow.PassedArgs`, you can discover this yourself if you simply don't use `PassthroughFlowRepresentable`. 

I can try to fix that as part of this issue, but it feels like this'll get stuck on behalf of a bug that has nothing to do with this new feature for a while.

Tyler-Keith-Thompson:
Code that is the problem, in AnyFlowRepresentable.swift
```swift
/**
     Creates an erased `FlowRepresentable` by using its initializer
     - Parameter type: The `FlowRepresentable` type to create an instance of
     - Parameter args: The `AnyWorkflow.PassedArgs` to create the instance with. This ends up being cast into the `FlowRepresentable.WorkflowInput`.
     */
    public init<FR: FlowRepresentable>(_ type: FR.Type, args: AnyWorkflow.PassedArgs) {
        argsHolder = args
        switch args {
            case _ where FR.WorkflowInput.self == Never.self:
                var instance = FR._factory(FR.self)
                _storage = AnyFlowRepresentableStorage(&instance)
            case _ where FR.WorkflowInput.self == AnyWorkflow.PassedArgs.self:
                // swiftlint:disable:next force_cast
                var instance = FR(with: args as! FR.WorkflowInput) // <-- THIS IS WRONG
                _storage = AnyFlowRepresentableStorage(&instance)
            case .args(let extracted):
                guard let cast = extracted as? FR.WorkflowInput else { fatalError("TYPE MISMATCH: \(String(describing: args)) is not type: \(FR.WorkflowInput.self)") }
                var instance = FR._factory(FR.self, with: cast)
                _storage = AnyFlowRepresentableStorage(&instance)
            default: fatalError("No arguments were passed to representable: \(FR.self), but it expected: \(FR.WorkflowInput.self)")
        }
        _storage._workflowPointer = self
        isPassthrough = FR.self is _PassthroughIdentifiable.Type
    }
```

Richard-Gist:
Wow that's not great.  Thank you for tracking that down.  I've already created an Issue to fix that particular bug from your comment.

Richard-Gist (approved):
> Having PassthroughFlowRepresentable work with StoryboardLoadable is going to be a requirement for me. Without it, I can't use it in our example app to showcase a simple way to leverage it.

Since you tracked this down to a pre-existing bug that is now tracked with issue #80 , I'm good with what I've reviewed.

morganzellers (approved):

wiemerm (approved):

Richard-Gist:
@wiemerm When you're good with this could ramp the minor version?  This enhances the API so we should do a minor bump, but it's not a breaking change so no need for a major bump.

Richard-Gist (commented):
Thank you @wiemerm

wiemerm (approved):

Richard-Gist (approved):

PR #88, opened 2021-07-28
Adds debugging to help diagnose flakey test

Richard-Gist:
## Checklist:


Richard-Gist:
EOD: I don't know if there is more that we can do at this time.  I will flood the pipeline periodically, but we may be back in a wait and see mode and hopefully have better debugging information.

Richard-Gist:
Success in reproducing!  https://github.com/wwt/SwiftCurrent/actions/runs/1076536741 has a failure with the new logs.

Richard-Gist:
REAL EOD: So we put in a class setup step.  If we don't see any test failures hooray, but also we have to dig through all of the logs to see if we were in the correct state during setUp.

morganzellers (approved):

wiemerm (approved):

PR #91, opened 2021-08-03
Fixes issue with changing parent view state and vanishing Workflow Views

### LINKED ISSUE #89: Abandoning and relaunching workflows in SwiftUI
### Describe the bug
Your template says a clear and concise description of the problem is necessary but I cannot deliver on that, so here is a rambling and difficult explanation of the problem:

Using our sample app I am able to consistently reproduce an interesting bug. Essentially by chaining workflows I am able to make the profile screen disappear, LIKE MAGIC! The branch `meetup` has this code reproduced. Short version: The profile screen has a workflow that `isLaunched: .constant(true)` and it abandons that workflow to logout.

While that part works beautifully then you proceed through your workflow again and after you login the profile screen just vanishes.

Debugging suggested to me that a fix might be finding a way (somehow...) to make the workflow a `@State`. I think that'll cause a host of other problems I don't want to think about, but I believe that is what needs to be done. This is more of a gut feeling but it's because instances of workflow just keep changing. 

## To Reproduce
1. Go to profile
2. Log in
3. Log out
4. Log in
5. See error

### Expected behavior
It should not do that.

### Screenshots
https://user-images.githubusercontent.com/33705774/127522370-94335124-d561-4b3e-997c-248b2f6dd4a4.mp4

Tyler-Keith-Thompson:
## Checklist:

### DIFF
--- a/Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
+++ b/Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
@@ -627,4 +627,56 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
 
         wait(for: [expectOnFinish, expectViewLoaded], timeout: TestConstant.timeout)
     }
+
+    func testWorkflowCorrectlyHandlesState() throws {
+        final class Launcher: ObservableObject {
+            var onFinishCalled = false
+            init(workflow: AnyWorkflow,
+                 responder: OrchestrationResponder,
+                 launchArgs: AnyWorkflow.PassedArgs) {
+                if workflow.orchestrationResponder == nil {
+                    workflow.launch(withOrchestrationResponder: responder, passedArgs: launchArgs)
+                }
+            }
+        }
+        struct ModifiedWorkflowViewExplorer<A, W, C>: View, Inspectable {
+            @Binding var isLaunched: Bool
+
+            let inspection = Inspection<Self>()
+            @State var wrapped: W?
+            @State var workflow: AnyWorkflow
+            @State var launchArgs: AnyWorkflow.PassedArgs
+            @State var onFinish = [(AnyWorkflow.PassedArgs) -> Void]()
+            @State var onAbandon = [() -> Void]()
+
+            @StateObject private var model: WorkflowViewModel
+            @StateObject private var launcher: Launcher
+
+            var body: some View { EmptyView() }
+        }
+
+        struct FR1: View, FlowRepresentable {
+            weak var _workflowPointer: AnyFlowRepresentable?
+
+            var body: some View {
+                Button("Proceed") { proceedInWorkflow() }
+            }
+        }
+
+        struct FR2: View, FlowRepresentable {
+            weak var _workflowPointer: AnyFlowRepresentable?
+
+            var body: some View {
+                Button("Abandon") { workflow?.abandon() }
+            }
+        }
+
+        let workflowView = WorkflowLauncher(isLaunched: .constant(true))
+            .thenProceed(with: WorkflowItem(FR1.self))
+
+        let explorer = unsafeBitCast(workflowView, to: ModifiedWorkflowViewExplorer<Never, Never, FR1>.self)
+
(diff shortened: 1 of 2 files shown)


morganzellers (approved):
I support these comments staying in 😅

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -627,4 +627,48 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
...
+            }
+        }
+
+        let workflowView = WorkflowLauncher(isLaunched: .constant(true))
+            .thenProceed(with: WorkflowItem(FR1.self))
+
+        let explorer = unsafeBitCast(workflowView, to: ModifiedWorkflowViewExplorer<Never, Never, FR1>.self)
+
+        // if the unsafeBitCast worked, then everything is a state var.
+        XCTAssertNil(explorer.wrapped)
```
Richard-Gist, line 672: This is true, but also, the bit cast could fail if we reorder the parameters, right?  Does it throw, or is explorer nil?

Also, why do we want `explorer.wrapped` to be nil?

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -627,4 +627,48 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
...
+            @Binding var isLaunched: Bool
+
+            let inspection = Inspection<Self>()
+            @State var wrapped: W?
+            @State var workflow: AnyWorkflow
+            @State var launchArgs: AnyWorkflow.PassedArgs
+            @State var onFinish = [(AnyWorkflow.PassedArgs) -> Void]()
+            @State var onAbandon = [() -> Void]()
+
+            @StateObject private var model: WorkflowViewModel
```
Richard-Gist, line 652: I see these are `private` but the others were made `internal`, but only `wrapped` is accessed?  Any reason for this?

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: No, re-ordering of parameters is relatively unimportant unless those parameters are of the exact same type (like `State<String>`. Even that would only matter if we asserted on their contents. You can re-order properties in either struct as much as you want without inherently breaking the bitcast.

Tyler-Keith-Thompson: `explorer.wrapped` is asserted to be nil, so we read from one of the values. The funny thing about unsafe bit casting is that it's unsafe. So it's possible, although highly unlikely, that somebody writes a struct with MORE properties that takes up the same space in memory. If that happens, when we try to read from a property we could underflow or overflow leading to undefined behavior, which is generally bad. This test will at least expose that and probably blow up spectacularly, should that happen.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: In the tests? It really makes no difference. `private` or `public` or `internal` don't change the space they take up in memory. Because this isn't production code it really has no bearing. Any `private` things that are left are just cruft from a copy/paste. However, things that are internal are internal so that you *could* assert on them.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: Oh, final answer. If the bit cast fails (which it was when I wrote the test, cause TDD) it fatal errors.

Richard-Gist (approved):

wiemerm (approved):

PR #99, opened 2021-08-08
Adds xcframework files to our git releases

Tyler-Keith-Thompson:
## Checklist:

### DIFF
--- a/.github/build-xcframework.sh
+++ b/.github/build-xcframework.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # CREDIT: https://forums.swift.org/t/how-to-build-swift-package-as-xcframework/41414/4
-# NOTE: This does modify the Package.swift file, which is less than ideal. 
+# NOTE: This does modify the Package.swift file, which is less than ideal.
 set -x
 set -e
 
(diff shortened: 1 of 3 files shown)


Tyler-Keith-Thompson:
GitHub has archived their own action for creating releases. They recommended a different one which I switched to in this PR: https://github.com/actions/create-release

morganzellers (approved):

.github/build-xcframework.sh
```diff
@@ -0,0 +1,81 @@
...
+    RELEASE_FOLDER="Release-tvossimulator"
+    ;;
+    "macOS")
+    RELEASE_FOLDER="Release-macos"
+    ;;
+    esac
+
+    ARCHIVE_PATH=$RELEASE_FOLDER
+
+    # Rewrite Package.swift so that it declaras dynamic libraries, since the approach does not work with static libraries
```
wiemerm, line 39:
```suggestion
    # Rewrite Package.swift so that it declares dynamic libraries, since the approach does not work with static libraries
```

.github/build-xcframework.sh
```diff
@@ -0,0 +1,79 @@
...
+    RELEASE_FOLDER="Release-watchossimulator"
+    ;;
+  "TvOS")
+    RELEASE_FOLDER="Release-tvos"
+    ;;
+  "TvOS Simulator")
+    RELEASE_FOLDER="Release-tvossimulator"
+    ;;
+  "macOS")
+    RELEASE_FOLDER="Release-macos"
```
Richard-Gist, line 33: I tried to pull all 3 libraries into a Mac Catalyst app, and they all failed because `While building for Mac Catalyst, no library for this platform was found in...`

Tyler-Keith-Thompson: Is this something you'd like me to fix before you'll consider merging or is this informational and something we'll need to add in as a follow-up?

Richard-Gist: I'm checking to see if it's something we can tweak right now or if it will need to be a follow-up.  Leaving the comment to prevent merging until we decide which way to go.

Richard-Gist: I think we'll wait until a request comes in for Catalyst in the hosted frameworks.  We know we support it using SPM, and this is just meant to be an additional option, not the primary form of consuming our library.

Richard-Gist (approved):

morganzellers (approved):

PR #104, opened 2021-08-11
Adds special thanks section to README that recognizes test dependencies

morganzellers:
## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ If you like what you've seen, consider [giving us a star](https://github.com/wwt
 
 # Special Thanks
 
-SwiftCurrent would not be nearly as fun to work on without all of the great work done on our test dependencies:
+SwiftCurrent would not be nearly as amazing without all of the great work done by the authors of our test dependencies:
 
 - [CwlCatchException](https://github.com/mattgallagher/CwlCatchException)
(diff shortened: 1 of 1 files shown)


morganzellers:
Draft only until the verbiage-riffing (if any) has subsided

README.md
```diff
@@ -127,6 +127,10 @@ And just like that you're started!
 - [Upgrade Path](https://github.com/wwt/SwiftCurrent/blob/main/wiki/UPGRADE_PATH.md)
 - [Contributing to SwiftCurrent](https://github.com/wwt/SwiftCurrent/blob/main/.github/CONTRIBUTING.md)
 
+# Special Thanks
```
Tyler-Keith-Thompson, line 130: My $0.02 is that we should move this below feedback. Also, why not just go ahead and call out the other dependencies?

Tyler-Keith-Thompson: Consider making it clear these are test-only dependencies. It's a bonus for us that people recognize we do not have any production dependencies outside of Apple 1st party frameworks.

morganzellers: I like the suggestion to include the others and specify as test-only. I had the section below feedback at first and I'm thinking I like that placement more as well.  - will update!

Tyler-Keith-Thompson (approved):

wiemerm (approved):

PR #106, opened 2021-08-11
Adding section regarding commit message format to CONTRIBUTING.md.

morganzellers:
## Checklist:

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -17,6 +17,10 @@ If you have an idea or change you would like to request, please go to [issues](h
 ## Pull Requests
 PRs have a checklist that walks you through the things that you should do for any PR. This document provides greater detail and context around what some of those checklist items mean. Please take the time to read through this document and our styleguide before contributing so that we can keep SwiftCurrent clean, and maintainable. By the time you submit code we want to be focused on the logic, not the formatting, or patterns. Also please be sure to read the section on testing in this guide as tests are incredibly important to us.
 
+> The SwiftCurrent team finds value in structuring our commit messages in a consistent way. We would love it if you would do the same:
+> 
+> ```[branch-name] - description```
+
 ## Sign the CLA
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -17,6 +17,10 @@ If you have an idea or change you would like to request, please go to [issues](h
 ## Pull Requests
 PRs have a checklist that walks you through the things that you should do for any PR. This document provides greater detail and context around what some of those checklist items mean. Please take the time to read through this document and our styleguide before contributing so that we can keep SwiftCurrent clean, and maintainable. By the time you submit code we want to be focused on the logic, not the formatting, or patterns. Also please be sure to read the section on testing in this guide as tests are incredibly important to us.
 
+> The SwiftCurrent team finds value in structuring our commit messages in a consistent way. We would love it if you would do the same:
+> 
+> ```[branch-name] - description```
+
```
wiemerm, line 23: A few questions around this:

1. Do we find enough value in consistent commit messages as to call it out?
2. If someone submits a PR that does not follow this convention are we going to reject the PR and ask them to amend their commits to follow this format? I'd caution against that as an additional barrier to contribute to the project.
3. Does it make sense to either include initials as part of a preferred commit message, or more so to remove the initials from our commits.

wiemerm: So to 1 & 2 how do we best word this to make it clear that we would like this format but do not require it? As for 3, I don't entirely agree. I view initials as typically just a note of "here's who I could go to for questions about this commit" or "here's who wrote that," but the Co-authored provides that already (and much more effectively) leaving the initials as redundant. This is a larger team question more of preference than consequence; still want to pose the question.

brianlombardo (approved):

Richard-Gist (approved):
I'll let @wiemerm resolve their conversation, but this looks good to me.

.github/CONTRIBUTING.md
wiemerm: Team discussed and we will be keeping initials.

PR #107, opened 2021-08-12
Changes to support library development on an M1 mac

Richard-Gist:
## Checklist:

### DIFF
--- a/Tests/SwiftCurrentTests/CustomAssertions.swift
+++ b/Tests/SwiftCurrentTests/CustomAssertions.swift
@@ -11,7 +11,8 @@ import XCTest
 
 @testable import CwlPreconditionTesting
 
-func XCTAssertThrowsFatalError(instructions: @escaping () -> Void, file: StaticString = #file, line: UInt = #line) {
+func XCTAssertThrowsFatalError(instructions: @escaping () -> Void, file: StaticString = #file, line: UInt = #line) throws {
+    #if (os(macOS) || os(iOS)) && arch(x86_64)
     var reached = false
     let exception = catchBadInstruction {
         instructions()
@@ -19,4 +20,7 @@ func XCTAssertThrowsFatalError(instructions: @escaping () -> Void, file: StaticS
     }
     XCTAssertNotNil(exception, "No fatal error thrown", file: file, line: line)
     XCTAssertFalse(reached, "Code executed past expected fatal error", file: file, line: line)
+    #else
+    throw XCTSkip("XCTAssertThrowsFatalError is only available on macOS/iOS on x86_64 architecture.")
+    #endif
 }
--- a/ExampleApps/UIKitExample/SwiftCurrent_UIKitTests/Storyboard/StoryboardLoadableTests.swift
+++ b/ExampleApps/UIKitExample/SwiftCurrent_UIKitTests/Storyboard/StoryboardLoadableTests.swift
@@ -14,7 +14,7 @@ import SwiftCurrent
 @testable import SwiftCurrent_UIKit
 
 class StoryboardLoadableTests: XCTestCase {
-    func testStoryboardLoadableThrowsErrorWhenYouCallDefaultInitImplementation() {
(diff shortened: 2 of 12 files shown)


Richard-Gist:
According to this issue: https://github.com/mattgallagher/CwlPreconditionTesting/issues/21 we are unable to use CwlPreconditionTesting on ARM based devices (like the M1 macs).

Tests/SwiftCurrentTests/CustomAssertions.swift
```diff
@@ -11,12 +11,16 @@ import XCTest
...
     let exception = catchBadInstruction {
         instructions()
         reached = true
     }
     XCTAssertNotNil(exception, "No fatal error thrown", file: file, line: line)
     XCTAssertFalse(reached, "Code executed past expected fatal error", file: file, line: line)
+    #else
+    throw XCTSkip("XCTAssertThrowsFatalError is only available on macOS/iOS on x86_64 architecture.")
+    #endif
 }
```
Tyler-Keith-Thompson, line 26: I think we can be a little more explicit and a lot more flexible with something like this.

```suggestion
func XCTAssertThrowsFatalError(instructions: @escaping () -> Void, file: StaticString = #file, line: UInt = #line) throws {
    try XCTSkipUnless(ProcessInfo.canCatchBadInstruction, "XCTAssertThrowsFatalError is only available on macOS/iOS on x86_64 architecture.")
    var reached = false
    let exception = catchBadInstruction {
        instructions()
        reached = true
    }
    XCTAssertNotNil(exception, "No fatal error thrown", file: file, line: line)
    XCTAssertFalse(reached, "Code executed past expected fatal error", file: file, line: line)
}

extension ProcessInfo {
    static var isx86: Bool {
        #if arch(x86_64)
            return true
        #else
            return false
        #endif
    }

    static var ismacOS: Bool {
        #if os(macOS)
            return true
        #else
            return false
        #endif
    }

    static var isiOS: Bool {
        #if os(iOS)
            return true
        #else
            return false
        #endif
    }

    fileprivate static var canCatchBadInstruction: Bool {
        (ismacOS || isiOS) && isx86
    }
}
```

Tyler-Keith-Thompson: As an added benefit making more things runtime like this means it's harder for somebody to accidentally break stuff when running on a particular architecture.

Tyler-Keith-Thompson: To prove out my previous statement on your M1 mac go delete a character from inside the `#if` but before the `#else`. It'll still happily compile for you, but not anybody running x86.

Richard-Gist: The issue that mine is solving is that it's a compile-time error.  My computer doesn't have `catchBadInstruction` to even reference, so I can't compile.  Otherwise, I would agree that not borking the other architectures would be ideal.

Tyler-Keith-Thompson (approved):

Richard-Gist:
There was one more change I had to do, but I believe it is on the person who has an M1 to do.  SwiftLint from the command line failed because something was misconfigured.  A quick google search later got me fixed up with the solution in this issue: https://github.com/realm/SwiftLint/issues/1466

morganzellers (approved):

brennanMKE:
Thanks for this useful code. I copied it to a standalone project so that I could work out some details to make the build settings work as a package used by multiple test targets. I needed it to prepare this [PR](https://github.com/aws-amplify/amplify-ios/pull/1548) to support Apple Silicon.

* [PreconditionCatching](https://github.com/brennanMKE/PreconditionCatching)

brennanMKE:
@Richard-Gist I packaged up these functions to make them easy to use as a dependency. It currently points to a forked branch which has the fix thanks for Saagar. I also have these functions limited to iOS and macOS since the underlying library only includes those 2 platforms. For other platforms it just skips them.

* [Test Assertions](https://github.com/brennanMKE/Hela/blob/main/Sources/Hela/TestAssertions.swift)

PR #116, opened 2021-08-29
NavigationLinks are now supported in SwiftUI!

Tyler-Keith-Thompson:
## Checklist:

----

### If Applicable:


Tyler-Keith-Thompson:
As discussed I am deliberately working on just writing failing tests, that leaves the team to focus on production code. As of the time of writing there are more tests that need to be written but I am quite confident in the testing suite, I am also confident in our ability to get nav stacks working along with skipping one or more items, `persistWhenSkipped` persistence settings, and adding `embedInNavigationView()` to the API surface. 

I suspect `removedAfterProceeding` will never animate the way we want it to, although I could be wrong. In UIKit this clearly means the next item slides in and the one you were just on gets removed from the back stack. While I am not at all confident in our ability to animate that the same way I do feel there's some middle ground where we just do it, and let SwiftUI animate that however it thinks is right. If that behavior is odd, or unlikely to be used that's fine, we're responding to a request from the consumer to remove the item after proceeding, and that is what we do. 

NOTE: removedAfterProceeding may just outright work differently (see: very likely will) between Xcode 12.4, 12.5 and 13. SwiftUI changed a fair amount between all of those versions. 

Alternative approaches for removedAfterProceeding:
- Don't support it by doing some complex nonsense to prevent that as a possibility at compile time (I hate this for maintenance reasons, and because I do not think we will get the compiler error messages we'd want to make this nice)
- Make setting the persistence a throwing function. On the surface this seems nice but in practice I suspect it will just make the consumer experience poor when trying to create views. Especially given that most of the time it won't fail.
- Create a runtime crash if you have a presentationType of `navigationLink` and a persistence of `removedAfterProceeding`. This is sub-optimal because it's a runtime issue. That said it's not an option I'm substantially opposed to. I had originally thought that data driven workflows might make this hard-to-spot but data driven workflows will already have linting, so we could dis-allow the data model setting that presentation type and that persistence.

Sources/SwiftCurrent_SwiftUI/Views/NavigationWrapper.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+//  Created by Tyler Thompson on 8/27/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+extension View {
+    func navLink<D: View>(to destination: D, isActive: Binding<Bool>) -> some View {
```
Tyler-Keith-Thompson, line 13: I left this in here, it's the workaround we had to use for nav links that are defaulted to true.

Tests/SwiftCurrent_SwiftUITests/AnyFlowRepresentableViewTests.swift
```diff
@@ -15,6 +15,13 @@ import SwiftCurrent
 
 @available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
 final class AnyFlowRepresentableViewTests: XCTestCase, View {
+    override func tearDownWithError() throws {
```
Tyler-Keith-Thompson, line 18: This is my attempt to solve an issue we had. Short version: When you nested `inspect` calls too deep we weren't waiting for all of them. So I've set up this queue of expectations to wait on. This solution does work but comes with the drawback of having to put it in front of every test that uses the inspection extensions on `WorkflowItem`. 

I might've been able to hack something together that just did it for all tests, but that would make debugging harder.

SwiftCurrent.xcworkspace/xcshareddata/xcschemes/SwiftCurrent_SwiftUI.xcscheme
```diff
@@ -52,6 +52,11 @@
                BlueprintName = "SwiftCurrent-SwiftUITests"
                ReferencedContainer = "container:">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "GenericConstraintTests/testProceedingWhenInputIsAnyWorkflowPassedArgsWithClosureFlowPersistence_WorkflowCanProceedToNeverItem()">
```
Richard-Gist: Why is this one being skipped?

Tests/SwiftCurrent_SwiftUITests/AnyFlowRepresentableViewTests.swift
Richard-Gist: We could extract it into an extension on XCTestCase that is called in the `tearDownWithError()`.  This will decrease the duplication but still allow each TestCase to decide if it will do this.

morganzellers (dismissed):

.github/.jazzy.yaml
```diff
@@ -7,4 +7,47 @@ module: SwiftCurrent
...
+    - FlowRepresentable
+    - PassthroughFlowRepresentable
+    - Workflow
+- name: Creating Workflows in SwiftUI
+  children:
+    - View
+    - ViewControllerWrapper
+    - WorkflowItem
+    - WorkflowLauncher
+- name: Creating Workflows in UIKit
```
Richard-Gist, line 33: I like how there is a little preamble for those who click `How to use SwiftCurrent with SwiftUI` and I think we may want something here as well.

Tyler-Keith-Thompson: Agreed, think we need a documentation follow-up.

Richard-Gist: I agree, and also, I can't GitHub.  I meant to put this on `Creating Workflows in SwiftUI` specifically because this did some shuffling of what you care about and how it can be used.

.github/abstract/How to use SwiftCurrent with SwiftUI.md
```diff
@@ -0,0 +1 @@
+SwiftCurrent has first-class SwiftUI support. It’s designed to naturally feel like using any other SwiftUI tool. It is optimized to avoid common pitfalls. For example, SwiftCurrent does not use `AnyView` in its implementation. This means all of the animation and performance benefits you get when using SwiftUI normally you get when using SwiftCurrent.
```
Richard-Gist, line 1: The last sentence caused me to stumble a little bit.  I think this is better for now.
```suggestion
SwiftCurrent has first-class SwiftUI support. It’s designed to naturally feel like using any other SwiftUI tool. It is optimized to avoid common pitfalls. For example, SwiftCurrent does not use `AnyView` in its implementation. This means that, when using SwiftCurrent, you will keep all the animation and performance benefits you would normally have gotten.
```

.github/wiki/Getting-Started-Storyboard.md
```diff
@@ -243,3 +243,26 @@ At a high level we are loading the view controller for testing (similar to befor
 <details>
```
Richard-Gist, line 243: Now that these are in the Jazzy docs.  I'm not sure we still need the wiki files anymore.  We could delete this file, right?

Tyler-Keith-Thompson: Not as part of this GET YOUR OWN PR FOR DOCUMENTATION STUFF

Richard-Gist (approved):
I'm going to follow this up with a bigger documentation PR, but for an initial BETA release, this is fine.

PR #117, opened 2021-08-31
GitHub io documentation restructured.

Richard-Gist:
## Checklist:
- [X] Did you sign the [Contributor License Agreement](https://cla-assistant.io/wwt/SwiftCurrent)?
- [X] Did you [update the documentation](https://github.com/wwt/SwiftCurrent/blob/main/.github/CONTRIBUTING.md#documentation)?


.github/.jazzy.yaml
```diff
@@ -11,31 +11,34 @@ author: WWT and Tyler Thompson
...
 - name: Creating Workflows in SwiftUI
   children:
+    - WorkflowLauncher
     - View
-    - ViewControllerWrapper
     - WorkflowItem
-    - WorkflowLauncher
+- name: How to use SwiftCurrent with UIKit
+  children:
+    - Getting Started with Programmatic Views
```
Tyler-Keith-Thompson, line 34: This is getting cut off. It says "Getting Started with Programmatic V". That's less than ideal, are we just accepting that for now?

README.md
```diff
@@ -1,9 +1,9 @@
- ![SwiftCurrent](https://raw.githubusercontent.com/wwt/SwiftCurrent/main/.github/wiki/swiftcurrent-logo.png)
+![SwiftCurrent](https://user-images.githubusercontent.com/79471462/131564417-6f4976f4-270c-41b3-bbe1-428528e2cc2c.png)
 
 <!-- Library Information -->
 [![Supported Platforms](https://img.shields.io/cocoapods/p/SwiftCurrent)](https://github.com/wwt/SwiftCurrent/security/policy)
-[![Swift Package Manager](https://img.shields.io/badge/Swift_Package_Manager-supported-brightgreen)](https://github.com/wwt/SwiftCurrent/wiki/Installation#swift-package-manager)
-[![Pod Version](https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout)](https://github.com/wwt/SwiftCurrent/wiki/Installation#cocoapods)
+[![Swift Package Manager](https://img.shields.io/badge/Swift_Package_Manager-supported-brightgreen)](https://wwt.github.io/SwiftCurrent/installation.html#swift-package-manager)
```
Tyler-Keith-Thompson, line 5: Is this actually supposed to be an io link? This looks like it could've been a bad find/replace. Same with the other one. Cause I thought badges needed github.com not the .io links

.github/.jazzy.yaml
Richard-Gist: I'm inclined to accept that issue for this PR and have a follow-up PR to hash out Abstracts and names.  But I'm also open to another rename.  I would want it to read well for all 3 Getting Started.  I played with:
```suggestion
    - Getting Started Programmatically
```
But that didn't read as well in this instance, nor did it translate to the other 2.

We can start a thread here for other names we could use.

README.md
Richard-Gist: Yes, it is.  This badge is generated with the first link

```markdown
![Swift Package Manager](https://img.shields.io/badge/Swift_Package_Manager-supported-brightgreen)
```

Then we have added a link to the badge so that on clicking it, it will take you to something that makes sense.  We had been redirecting to the installation page which is what this change is continuing.

morganzellers (dismissed):

.github/.jazzy.yaml
Tyler-Keith-Thompson: The issue is that these have to be globally unique, or at least I could not see anything from Jazzy that let you map file names to headers. So that tends to make us want to be more verbose.

I really wish our docs didn't cut off. Have you considered:
```suggestion
- name: How to use SwiftCurrent with UIKit
  children:
    - With Programmatic Views
    - With Storyboards
```

morganzellers: I like how the UIKit section looks with children

Richard-Gist: I can see that looking pretty good on the sidebar.  I'm not sure for SwiftUI though, let's see:

```
- name: How to use SwiftCurrent with SwiftUI
  children:
    - With SwiftUI
```
Hmm.... What if we did something like:

```
- name: How to use SwiftCurrent with UIKit
  children:
    - Using Programmatic Views
    - Using Storyboards
```

Then SwiftUI might look like this:
```
- name: How to use SwiftCurrent with SwiftUI
  children:
    - Using View Swapping
```

Tyler-Keith-Thompson: I sort of hate "Using View Swapping" but I really like everything else.

Richard-Gist: Is it the calling out that it uses "View Swapping" or that there is terminology better suited for swapping out the views?

Or option 3, you hate it because it's the only in the list?  And if we had:
```
  - Using View Swapping
  - Using NavigationViews
```

Would be better?  Though those would link to wildly different articles...

Tyler-Keith-Thompson: I sort of didn't like View Swapping as a term.

Richard-Gist: Ok so:
* Rename UIKit articles to Using
* Leave SwiftUI the way it is

Then we'll get this PR merged and come back to more naming/Abstracts later?

PR #118, opened 2021-08-31
Now supporting sheets in SwiftUI!

morganzellers:
## Checklist:

----

### If Applicable:


Sources/SwiftCurrent_SwiftUI/ViewInspector/InspectableSheet.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+//  Created by Morgan Zellers on 8/31/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//
+// swiftlint:disable file_types_order
+import SwiftUI
+
+@available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
+extension View {
+    func testableSheet<Sheet>(isPresented: Binding<Bool>, onDismiss: (() -> Void)? = nil, @ViewBuilder content: @escaping () -> Sheet
```
Richard-Gist, line 13: I get why we do this.  And I realize the name is pretty self-explanatory, but I wonder if we should still add a little more documentation to this?  I'm conflicted, so feel free to say no.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
```diff
@@ -49,7 +49,9 @@ public struct WorkflowItem<F: FlowRepresentable & View, Wrapped: View, Content:
         ViewBuilder {
             if launchStyle == .navigationLink, let content = content {
                 content.navLink(to: nextView, isActive: $isActive)
-            } else if let body = model.body?.extractErasedView() as? Content, elementRef == nil || elementRef === model.body, launchStyle == .default {
+            } else if (wrapped as? WorkflowItemPresentable)?.workflowLaunchStyle == .modal, let content = content {
+                content.testableSheet(isPresented: $isActive) { nextView }
```
Richard-Gist, line 53: There might be a better name for `$isActive` but I'm not sure what it is yet.  When I first read this, it always takes me a minute to realize what `isActive` is referring to (the next screen being active/this link being active, NOT the current screen being active).

Sources/SwiftCurrent_SwiftUI/Extensions/SwiftUILaunchStyleAdditions.swift
```diff
@@ -44,6 +51,7 @@ extension LaunchStyle {
             public var rawValue: LaunchStyle {
                 switch self {
                     case .navigationLink: return ._swiftUI_navigationLink
+                    case .modal: return ._swiftUI_modal
```
Richard-Gist, line 54: We'll see how this PR ends up, but I'm wondering if it makes sense to make `.modal` function similarly as it does on UIKit where it takes in its own enumeration to style the modal further?  I haven't played with Modals/sheets in SwiftUI enough to figure out all the customizations we may be hiding with a single enumeration.

Richard-Gist (commented):
Mostly questions and musings.  Everything looks good at this point.

Sources/SwiftCurrent_SwiftUI/ViewInspector/InspectableSheet.swift
Tyler-Keith-Thompson: No.

Tyler-Keith-Thompson: To expand on that: This isn't a public symbol, nor is it one that you have to worry about even if you're a consumer using ViewInspector doing an `@testable import SwiftCurrent_SwiftUI`. So unless it's somehow unclear to maintaners I think we're okay.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
```diff
@@ -49,7 +49,9 @@ public struct WorkflowItem<F: FlowRepresentable & View, Wrapped: View, Content:
...
+            } else if let body = model.body?.extractErasedView() as? Content, elementRef == nil || elementRef === model.body, launchStyle != .navigationLink {
```
Richard-Gist, line 54: I'm not entirely sure why `launchStyle != .navigationStyle`.  What scenario is that covering?

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
Tyler-Keith-Thompson: Hrm, that's what navigation links call it, modal call it `isPresented`. Both of those would cause your confusion. Then again with this highly recursive structure I'm not sure there's a very clear way to describe it.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
Tyler-Keith-Thompson: That was the *first* thing we ran into. So the short version is that if that goes back to `launchStyle == .default` then the last screen in a series of modals (or the only screen if there is just one modal) is blank. Because the final screen just needs to render the view. 

So if FR1 -> FR2 (modal) then FR2 (modal) doesn't wrap itself in a sheet, it just *is*. Which is what that whole `if let body` line is about (just present myself).

Sources/SwiftCurrent_SwiftUI/Extensions/SwiftUILaunchStyleAdditions.swift
Tyler-Keith-Thompson: I think we should do that in a follow-up PR. It should probably be the next thing we work on, though.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
Richard-Gist: Ok.  So if we added other non-modal launch styles, we would need to be careful that this continues to work properly (alerts?).  But that's a future hypothetical problem to deal with.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
Richard-Gist: What about: `launchStyleIsActive`?  Or `launchStyleModifierIsActivated`?

Probably should go with: `isThisEverGoingToEndRichard`

morganzellers: I'll +1 `launchStyleIsActive`

Tyler-Keith-Thompson: `launchStyleIsActive` is deceptive because the launch style is not the active thing. `launchStyleModifierIsActivated` is certainly verbose, but a poor descriptor as NavLinks are not modifiers, even if you're being incredibly pedantic about how we made a thing that *looks* kind of like a modifier for nav links it's not actually a modifier, it's just an extension function. 

`isThisEverGoingToEndRichard` probably doesn't give maintainers the opportunity they need to understand any better than `isActive` does.

Richard-Gist: I disagree. `isThisEverGoingToEndRichard` clearly points to who caused the problem and maintenance nightmare.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowItem.swift
Tyler-Keith-Thompson: That's accurate but I might describe that in terms of how nav links work vs how modifiers work. Nav links take a destination and wrap a view to present. Modals, alerts, etc are *modifiers* that have access to the current view, have access to a binding boolean or item, and then wrap the thing they'll be presenting. 

So that's my long-winded way of saying that it's not non-modal launch styles, it's simply if we add anything else that works like nav links do, we'll need to modify this conditional again. But alerts would actually still work under this current logic.

Tyler-Keith-Thompson (approved):

Richard-Gist (approved):

PR #121, opened 2021-09-01
Adds support for Full screen cover modal presentation style

Richard-Gist:
## Checklist:

### DIFF
--- a/.github/SwiftCurrentLint/.swiftlint.yml
+++ b/.github/SwiftCurrentLint/.swiftlint.yml
@@ -129,6 +129,7 @@ included: # paths to include during linting. `--path` is ignored if present.
   # - ../../Tests
 excluded: # paths to ignore during linting. Takes precedence over `included`.
   - ../../ExampleApps/SwiftUIExample/SwiftCurrent_SwiftUI_UITests
+  - ../../ExampleApps/SwiftUIExample/SwiftUIExample/TestViews
 #   - Workflow/.swiftpm/
 #   - Workflow/Package.swift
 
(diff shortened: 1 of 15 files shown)


Richard-Gist:
EOD: Because fullScreenCover is untestable by ViewInspector, we will be adding our first XCUI tests.  Additionally, `fullScreenCover` is not available in macOS, and I missed a step there that @Tyler-Keith-Thompson has probably already fixed before I can post this comment.

For the XCUI tests, you can see the last commits working towards an idea for how they will work.  Ultimately I think where we want to go with it is that there are about 4 FRs that are shared code between the UI tests and SwiftUIExample, and they allow customization by taking in some argument that determines their `shouldLoad`.  We then have a couple of views that are returned how FR1 is being returned now, but instead of just being a single FR1, they would return a view that has the `WorkflowLauncher` with the appropriate number of FR's and use the EnvironmentKey to customize aspects of the view appropriately.

This makes most things Enum driven to talk between the test and the app, and the code readily available between the two for development.  Hopefully, we'll end up with about 4 FR views we use for testing and hopefully only about 4 or so views holding `WorkflowLauncher`s. We can do all the rest of our customizations from the environment we launch the app with.

Tyler-Keith-Thompson:
EOD: We have XCUITests that work...they really are just horrible tests to have to write. We will stick to ViewInspector as much as possible. I am currently satisfied by the level of testing the XCUITests provide. However, until now they were not running in the pipeline. We may need to tweak them so that they *can* run in the pipeline because I might have gotten too fancy with ResultBuilders for Xcode 12.4...we'll see.

However, once our pipeline passes I am personally satisfied, let't ship out fullScreenCover.


NOTE: A SwiftUI bug:
We discovered a frustrating SwiftUI bug where calling `dismiss` on `presentation` causes a bad animation and just general wonkiness when the navigation view is a column based navigation view. I haven't the slightest idea *why* we are seeing that behavior, and I am not certain to what extent it is a problem. If people use `embedInNavigationView()` on `WorkflowLauncher` they'll be just fine. Xcode 13 is bound to change all of that anyways, so I think we should proceed forward regardless.

Tyler-Keith-Thompson (approved):

PR #122, opened 2021-09-02
Sets CI to only run on pushes to main branch

morganzellers:
## Checklist:


Richard-Gist (approved):
So with this, we should only run this in Main, and everyone else will get it running in their main, or when they specifically request a run through workflow_dispatch.  With our working methods, we will only run the tests once per commit instead of twice.  AND it is now halving the boxes we use on each commit when we have a draft PR open.

We've discussed all this and this is what we want going forward.

PR #133, opened 2021-09-15
Made Changes to Creating Workflows Section

courtneybeisner:
Proofread and made a few changes. Note that when compile time and run time are nouns, no hyphens. When they are adjectives, they are hyphenated.

## Checklist:

----

### If Applicable:

----


.github/abstract/Creating Workflows.md
```diff
@@ -1,12 +1,12 @@
...
+### Valid Workflow Sequences
+Workflows enforce (either at compile time or run time) that the sequence of `FlowRepresentables` is well-formed. For example, if a particular `FlowRepresentable` has a `FlowRepresentable.WorkflowInput` of `String`, but the previous item passes an `Int` forward, that workflow is malformed. A `FlowRepresentable` can declare a `FlowRepresentable.WorkflowInput` of `Never` if it does not take in data, `Any` if it can intake multiple types, or `AnyWorkflow.PassedArgs` if it can handle being passed no data, or data of any kind. Lastly, if your `FlowRepresentable.WorkflowInput` needs to take in multiple values, you can use a tuple.
```
Richard-Gist, line 8: I apparently can't English.  Why are these "run time" instead of "run-time" when down below we went from "runtime" to "run-time"?

.github/abstract/Creating Workflows.md
```diff
@@ -1,12 +1,12 @@
...
 
-In SwiftUI, the compiler was not efficient enough to give the same compile-time feedback on malformed workflows. When that safety was added, the compiler only allowed for small workflows to be created. To combat this, SwiftUI is heavily run-time influenced. When you create a `WorkflowLauncher`, the launcher performs a runtime check to guarantee the workflow is well-formed. This means that if you wanted to test your workflow was well-formed, all you have to do is instantiate a `WorkflowLauncher`.
\ No newline at end of file
+In SwiftUI, the compiler was not efficient enough to give the same compile-time feedback on malformed workflows. When that safety was added, the compiler only allowed for small workflows to be created. To combat this, SwiftUI is heavily run time-influenced. When you create a `WorkflowLauncher`, the launcher performs a run time check to guarantee the workflow is well-formed. This means that if you wanted to test your workflow was well-formed, all you have to do is instantiate a `WorkflowLauncher`.
```
Richard-Gist, line 12: I'm also confused here with "heavily run time-influenced" versus "heavily run-time influenced" versus "heavily runtime influenced".

.github/abstract/Creating Workflows.md
Richard-Gist:
```suggestion
Workflows enforce (either at compile-time or run-time) that the sequence of `FlowRepresentables` is well-formed. For example, if a particular `FlowRepresentable` has a `FlowRepresentable.WorkflowInput` of `String`, but the previous item passes an `Int` forward, that workflow is malformed. A `FlowRepresentable` can declare a `FlowRepresentable.WorkflowInput` of `Never` if it does not take in data, `Any` if it can intake multiple types, or `AnyWorkflow.PassedArgs` if it can handle being passed no data, or data of any kind. Lastly, if your `FlowRepresentable.WorkflowInput` needs to take in multiple values, you can use a tuple.
```

.github/abstract/Creating Workflows.md
Richard-Gist:
```suggestion
In SwiftUI, the compiler was not efficient enough to give the same compile-time feedback on malformed workflows. When that safety was added, the compiler only allowed for small workflows to be created. To combat this, SwiftUI is heavily run-time influenced. When you create a `WorkflowLauncher`, the launcher performs a run-time check to guarantee the workflow is well-formed. This means that if you wanted to test your workflow was well-formed, all you have to do is instantiate a `WorkflowLauncher`.
```

Richard-Gist (approved):

morganzellers (approved):

PR #145, opened 2021-09-24
Updates SwiftUI app to match the demo video

Richard-Gist:
## Checklist:
- [ ] Did you sign the [Contributor License Agreement](https://cla-assistant.io/wwt/SwiftCurrent)?

----

### If Applicable:

----


Richard-Gist:
I'm in process of making the demo app testable but I have hit 1 main roadblock.

1) The modifiers we're putting on the PasswordWorkflow make it uncastable in ViewInspector thus failing.

For 1, I love the flexibility that it showcases with embedding things in a NavigationView, so I'm reluctant to change the prod code for testability.

morganzellers (approved):

Tyler-Keith-Thompson (changes requested):
Needs better test coverage, and all the other things you are already planning. Writing this to prevent a merge since it just got approved.

Richard-Gist:
So we had a little chat and I'm going to reorient my approach to our example apps.

From now on, our example apps will focus on a production-worthy app that uses SwiftCurrent to achieve its goals.  The testing for these apps would reflect how we would do our own testing for a production app, moving away from testing specifically for SwiftCurrent.  In addition, because these would be prod apps now, we'll want to move out the hosting application for library testing into a new target to keep test code out of prod.

ExampleApps/UIKitExample/SwiftCurrent_UIKitTests/TestUtilities/NotificationReceiverLocal.swift
```diff
@@ -0,0 +1,55 @@
+//
+//  NotificationReceiverLocal.swift
```
Richard-Gist, line 2: This thing and things related to it (data, protocol, this receiver) should all move into `SwiftCurrent_Test` and be a part of the testing suite of tools.  I haven't done it yet, because I'm looking for feedback on more tweaks or value of this.  I think it's not complete but if we kept polishing, this would be a decent way of providing flexibility to people testing, while abstracting away the nitty-gritty of doing it.

There's probably too much overlap with the Local version and the Test version of this receive.

ExampleApps/UIKitExample/SwiftCurrent_UIKitTests/TestUtilities/NotificationReceiverLocal.swift
```diff
@@ -12,34 +12,19 @@ protocol WorkflowTestingReceiver {
     static var workflowTestingData: WorkflowTestingData? { get set }
 }
 
-// MARK: WORST TESTING WORKAROUND EVA!!!
+// MARK: Getting better but still doesn't feel ideal
```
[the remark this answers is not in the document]
Richard-Gist: Thanks.  As I've been looking into moving it into the Testing module, it's not yet ready for redistribution.  If we're cool with it, I think I'm going to leave it in Test Utilities for now, and then we'll do a different PR to make a test package version.

ExampleApps/SwiftUIExample/SwiftUIExampleTests/Views/LoginTests.swift
```diff
@@ -0,0 +1,94 @@
...
+        wait(for: [exp], timeout: TestConstant.timeout)
+
+        XCTAssertNotNil(loginView)
+
+        waitUntil(Self.workflowTestingData != nil)
+        let data = Self.workflowTestingData
+
+        // Test Workflow arrangement
+        XCTAssertEqual(data?.workflow.count, 2)
+        let first = data?.workflow.first { _ in true }
```
Tyler-Keith-Thompson, line 71: Workflow already has a `.first` property...why are we using a closure?

Tyler-Keith-Thompson: If it's an issue with AnyWorkflow it feels better to expose it than to do this.

ExampleApps/SwiftUIExample/SwiftUIExampleTests/Views/TermsAndConditionsTests.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+        wait(for: [exp], timeout: TestConstant.timeout)
+    }
+
+    func testPrimaryAcceptButtonCompletesWorkflow() {
+        let workflowFinished = expectation(description: "View Proceeded")
+        let exp = ViewHosting.loadView(WorkflowLauncher(isLaunched: .constant(true)) {
+            thenProceed(with: TermsAndConditions.self)
+        }.onAbandon {
+            XCTFail("Abandon should not have been called")
```
Tyler-Keith-Thompson, line 29: Was this part of a refactor? I generally am wary of testing things that didn't happen, cause where do you stop?

ExampleApps/SwiftUIExample/Views/Design/Colors.swift
```diff
@@ -0,0 +1,41 @@
...
+    static var primaryBackground: Color {
+        .black
+    }
+
+    static var primaryText: Color {
+        white
+    }
+
+    static var card: Color {
+        Color(red: 86 / 255, green: 86 / 255, blue: 118 / 255)
```
Tyler-Keith-Thompson, line 22: Thought we were gonna extract this into a new initializer that already did the / 255 thing.

ExampleApps/SwiftUIExample/Views/GenericOnboardingView.swift
```diff
@@ -0,0 +1,68 @@
...
+//  Created by Richard Gist on 9/22/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//  
+
+import SwiftUI
+import SwiftCurrent
+import Swinject
+
+struct GenericOnboardingView: View, FlowRepresentable {
+    @DependencyInjected private static var userDefaults: UserDefaults!
```
Tyler-Keith-Thompson, line 14: Why are we not using `@AppStorage` here?

ExampleApps/SwiftUIExample/Views/LoginView.swift
```diff
@@ -0,0 +1,101 @@
...
+                            .titleStyle()
+
+                        VStack(spacing: 15) {
+                            PrimaryTextField(placeholder: "Email Address", image: Image.account, text: $email)
+                                .keyboardType(.emailAddress)
+
+                            PasswordField(showPassword: $showPassword, password: $password)
+                        }
+
+                        HStack {
```
Tyler-Keith-Thompson, line 43: ? An HStack with only a spacer in it? I'm sure there's a reason for this but I can't fathom what it is.

ExampleApps/SwiftUIExample/Views/Profile/AccountInformationView.swift
```diff
@@ -12,42 +12,87 @@ import SwiftCurrent_SwiftUI
 
 struct AccountInformationView: View, FlowRepresentable {
     @State var password = "supersecure"
-    @State private var username = "changeme"
-    @State private var usernameWorkflowLaunched = false
-    @State private var passwordWorkflowLaunched = false
+    @State var email = "SwiftCurrent@wwt.com"
```
Tyler-Keith-Thompson, line 15: Do these need to be internal now?

ExampleApps/SwiftUIExample/Views/Profile/AccountInformationView.swift
```diff
@@ -12,42 +12,87 @@ import SwiftCurrent_SwiftUI
...
+    @State var emailWorkflowLaunched = false
+    @State var passwordWorkflowLaunched = false
```
Tyler-Keith-Thompson, line 17: Same question, do these need to be internal? Can they be private?

ExampleApps/SwiftUIExample/Views/Profile/ChangePasswordView.swift
```diff
@@ -11,9 +11,12 @@ import SwiftCurrent
 
 struct ChangePasswordView: View, FlowRepresentable {
     typealias WorkflowOutput = String
-    @State private var oldPassword: String = ""
-    @State private var newPassword: String = ""
-    @State private var confirmNewPassword: String = ""
+    @State private var oldPassword = ""
```
Tyler-Keith-Thompson, line 14: Minor nitpick, is there any way to create a data structure for this? A tuple, if nothing else, something so we don't have so many state variables that are just "Text for the thing, boolean for the thing"

PR #148, opened 2021-09-30
Docc

Tyler-Keith-Thompson:
## Checklist:


Package@swift-5.4.swift
```diff
@@ -0,0 +1,73 @@
+// swift-tools-version:5.3
```
Richard-Gist, line 1: This copy of Package will need to be deleted.

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.5
```
Richard-Gist, line 1: This change could be disruptive for consumers.  I'm ok with it but it needs to be more than a patch version bump.  Perhaps a feature or major version, but I'm not sure.

Something else we could do is create another project/package that is just for documentation and is not shipped to consumers.  Since all this documentation is for developers working on SwiftCurrent, we don't have to ship the project.  That project would build all things being documented (probably one reason the liblint fails for CocoaPods when adding more files from SwiftCurrent_SwiftUI)

Additional thought, I wonder if a dedicated documentation project would allow us to not alias every file we are documenting.

Richard-Gist (commented):
Dropping some comments to address before merging to main.  These are not my comments on the documentation itself (that'll be in a PR into this branch).

Package.swift
Tyler-Keith-Thompson: We spoke about this over a call, but the short version is that we do not want to create a separate project for consumers.

This change is *not* disruptive to consumers because of the other package.swift file you said we needed to delete...that's why we have 2 now.

Package@swift-5.4.swift
Tyler-Keith-Thompson: Nope, we have it there so that current consumers are not impacted by the version ramp.

Tyler-Keith-Thompson:
This was an interesting idea. It was an attempt to use DocC as a way of onboarding contributors to the library. I'm not sure that's a great use-case though. This can remain in our history if we ever want it back, but the PR isn't the biggest priority.

Tyler-Keith-Thompson:
This was an interesting idea. It was an attempt to use DocC as a way of onboarding contributors to the library. I'm not sure that's a great use-case though. This can remain in our history if we ever want it back, but the PR isn't the biggest priority.

PR #149, opened 2021-10-25
Updating trunk.yml for Testing framework

### LINKED ISSUE #110: XCFramework for SwiftCurrent_Testing
We've got other XCFrameworks for all of our packages except for SwiftCurrent_Testing. It would be good to add that as well, if we can.

morganzellers:
## Checklist:


Tyler-Keith-Thompson:
We should verify that this builds such that `@testable` actually works.

Tyler-Keith-Thompson:
Isn't the last thing to do on this just a quick verification that @testable works?

morganzellers:
> Isn't the last thing to do on this just a quick verification that @testable works?

No, if I remember correctly, the build command in `build-frameworks.sh` will need to be different for the testing package. I believe the last thing I was doing was trying to find the right build call and encountering the error above

morganzellers:
So this has moved forward some, but not all the way. The `build-xcframework.sh` script is still failing, I believe now, due to the wrong path being used to copy the modules or frameworks.

nickkaczmarek:
> This will not work as expected. The reason is we're using `xcodebuild` not `swift build` for creating the xcframework.
> 
> I've tried to look it up and was sorely disappointed: [developer.apple.com/forums/thread/125921](https://developer.apple.com/forums/thread/125921)

[This also seems related](https://developer.apple.com/forums/thread/650163), but similarly doesn't seem to provide the answer we want. It has some responses from DTS, but our use case might not be supported.

morganzellers:
It sounds like this is not something we will be following through on due to the `xcodebuild` / `swift build` differences. I plan to close this PR and the associated issue.

PR #150, opened 2021-11-03
Launching a Workflow using an AnyWorkflow

morganzellers:
## Checklist:

----

### If Applicable:

----


Richard-Gist (commented):
I'm going to try to stay hands-off, and let you run with this. Reach out if you want some help.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -731,6 +731,30 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase, App {
 
         wait(for: [exp], timeout: TestConstant.timeout)
     }
+
+    func testLaunchingAWorkflowFromAnAnyWorkflow() {
```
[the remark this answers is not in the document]
Tyler-Keith-Thompson: That's exactly what the generic constraint tests are for

morganzellers:
Added a test for fatalError when no items in a workflow earlier then spent most of the day deciphering some of the unit tests and working through adding needed `thenProceed`'s.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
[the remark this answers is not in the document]
morganzellers: @Tyler-Keith-Thompson I'll take a look at those in the AM 👍🏻

morganzellers:
EOD: Was able to drive out a few more `thenProceed`'s today. Starting to understand them more, but am still fuzzy at times on which situation in a workflow a thenProceed signature translates to. 

Also fuzzy on the difference between setting `typealias WorkflowOutput = String` vs `typealias WorkflowOutput = AnyWorkflow.PassedArgs`. I thought that setting the latter basically said "Whatever type I'm passed is what I'm passing on" but there were a few times today where I doubted my memory.

Sources/SwiftCurrent_SwiftUI/Extensions/WorkflowExtensions.swift
```diff
@@ -36,14 +36,40 @@ extension Workflow where F: FlowRepresentable & View {
...
+//        let workflow = Workflow<F>(first)
+//        workflow.append(ExtendedFlowRepresentableMetadata(flowRepresentableType: type,
+//                                                          launchStyle: launchStyle) { data in
+//            guard case.args(let extracted) = data,
+//                  let cast = extracted as? F.WorkflowInput else { return .default }
+//
+//            return flowPersistence(cast)
+//        })
+//        return workflow
+//    }
```
morganzellers, line 72: Hate to commit this much commented code, but I didn't trust myself to remember the signatures I want to drive out next over the weekend 😬.

morganzellers:
EOD: Added support for startingArgs - swapped to using the GenericConstraintsTests (copied and renamed for ExtendedFlowRepresentableMetadata) to drive out more functionality for persistence and launch style arguements.

Richard-Gist:
I tried to work on this a bit today, but there's a LOT of work we'll need to do to convert all of the tests over to actually test things.  I tried some different routes like maybe taking the generic constraint tests from SwiftCurrent core and seeing if those were easier to change over, but they weren't because of a difference in what you can use as starting arguments.

Maybe someone with a bit more regex prowess can make mass edits to convert the code, but for now, the only way forward I see is essentially translating all of the tests.  I found I had to update how the workflow was created, how it was launched, and how it was asserted.

I did contemplate for a while if we even needed to do all of this, which is still a "maybe" to a somewhat of a "yeah".  We might be able to prune it down some but a lot of these tests will remain valid if we ever get to a point of letting our users declare Workflows and feed them into SwiftUI this way (which is something that holds a lot of value for integrating into certain design patterns).

But overall I'm not sure we're actually going to use these `thenProceeds` to build out our workflows from data.

Sources/SwiftCurrent_SwiftUI/Views/WorkflowLauncher.swift
```diff
@@ -77,6 +77,38 @@ public struct WorkflowLauncher<Content: View>: View {
...
+        self.init(isLaunched: isLaunched, startingArgs: .none, workflow: AnyWorkflow(workflow))
+    }
+
+    /**
+     Creates a base for proceeding with a `WorkflowItem`.
+     - Parameter isLaunched: binding that controls launching the underlying `Workflow`.
+     - Parameter startingArgs: arguments passed to the first loaded `FlowRepresentable` in the underlying `Workflow`.
+     - Parameter workflow: workflow to be launched; must contain `FlowRepresentable`s of type `View`
+     */
+    public init<F: FlowRepresentable & View>(isLaunched: Binding<Bool>, startingArgs: AnyWorkflow.PassedArgs, workflow: Workflow<F>) where Content == AnyWorkflowItem {
```
Richard-Gist, line 95: We should rework this a little.  This is similar to the issue we just resolved.  `startingArgs` should have an iteration of `A` and an iteration of `where A == AnyWorkflow.PassedArgs`

Tests/SwiftCurrentTests/FlowRepresentableMetadataConsumerTests.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//  Created by Morgan Zellers on 11/2/21.
+//  Copyright © 2021 WWT and Tyler Thompson. All rights reserved.
+//  
+
+import XCTest
+import SwiftCurrent
+
+class FlowRepresentableMetadataConsumerTests: XCTestCase {
+
```
Richard-Gist, line 13:
```suggestion
```

Tests/SwiftCurrentTests/FlowRepresentableMetadataConsumerTests.swift
```diff
@@ -0,0 +1,44 @@
...
+
```
Richard-Gist, line 14:
```suggestion
```

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -731,6 +731,258 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase, App {
...
+            typealias WorkflowOutput = AnyWorkflow.PassedArgs
+            var _workflowPointer: AnyFlowRepresentable?
+            private let data: AnyWorkflow.PassedArgs
+            var body: some View { Text("FR2 type") }
+
+            init(with data: AnyWorkflow.PassedArgs) {
+                self.data = data
+            }
+        }
+//        struct FR3: View, FlowRepresentable, Inspectable {
```
Richard-Gist, line 770: Probably should remove?

Richard-Gist (approved):
Even though the Linter is failing, I am approving this PR because we are merging into a feature branch.  Those TODO failures will prevent us from accidentally merging into main with that code, but also allow us to move forward with getting data-driven working.

PR #152, opened 2021-11-09
Creates workflow to validate URLs link correctly.

Richard-Gist:
This should help protect SwiftCurrent from having deadlines in the README as well as the Github Pages site.

## Checklist:
- [X] Did you sign the [Contributor License Agreement](https://cla-assistant.io/wwt/SwiftCurrent)?


.github/workflows/validate_urls.yml
```diff
@@ -0,0 +1,39 @@
+name: Validate URLs
+
+on:
+  page_build:
+  schedule:
+    - cron: '00 14,23 * * *'
```
Tyler-Keith-Thompson, line 6: Suggestion: add a comment explaining the cron syntax. It's not intuitive and I wouldn't expect the majority of maintainers to get it right away

.github/workflows/validate_urls.yml
```diff
@@ -0,0 +1,39 @@
...
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to validate'
```
Tyler-Keith-Thompson, line 10: Can we make this description more useful?

.github/workflows/validate_urls.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  validate_documentation_links_on_branch:
+    name: Validate links on branch
+    runs-on: ubuntu-latest
+    env:
+      DEFAULT_BRANCH: main
+    steps:
+    - name: Configure variables
+      run: echo "GIT_BRANCH=${{ github.event.inputs.branch || env.DEFAULT_BRANCH }}" >> $GITHUB_ENV
+    - uses: urlstechie/urlchecker-action@master
```
Tyler-Keith-Thompson, line 23: We will need to add this to our accepted actions for SwiftCurrent. I will take that on shortly and comment again when it's done.

.github/workflows/validate_urls.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        git_path: https://github.com/wwt/SwiftCurrent
+        branch: ${{ env.GIT_BRANCH }}
+        file_types: .md,.swift
+        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new
```
Tyler-Keith-Thompson, line 28: Should there be additional excludes? Issues? Branches? We don't really need to validate GitHub things.

.github/workflows/validate_urls.yml
```diff
@@ -0,0 +1,39 @@
...
+        
+  validate_documentation_links_on_website:
+    name: Validate links on Github Pages
+    runs-on: ubuntu-latest
+    steps:
+    - uses: urlstechie/urlchecker-action@master
+      with:
+        git_path: https://github.com/wwt/SwiftCurrent
+        branch: gh-pages
+        file_types: .html
```
Tyler-Keith-Thompson, line 38: Is it worth validating the .md files here as well? I recognize they *should* all be transpiled into HTML by the time this runs, but it's unclear to me whether there's an edge case for .md files that we could cover.

Tyler-Keith-Thompson (commented):
Thank you for your PR! I'm excited that we'll be able to detect broken links, since that's bitten us before.

.github/workflows/validate_urls.yml
Richard-Gist: We can.  This is the name of the field when you manually dispatch the workflow.  Do you have any suggestions?

Richard-Gist: Screen Shot 2021-11-08 at 5 13 21 PM
Here is an example of what it would look like.

.github/workflows/validate_urls.yml
Richard-Gist: Added comments.  Not that everyone uses it, but the GitHub website editor gives you a hover comment to explain when it runs:
Screen Shot 2021-11-08 at 5 16 51 PM

.github/workflows/validate_urls.yml
Tyler-Keith-Thompson: Looks like this would *not* be for changing the branch from gh-pages to something else. 

"Name of branch to link crawl"

"Branch to crawl for broken links"

"Branch containing the markdown and swift you wish to crawl"

"Branch me, bro!"

Something like that (maybe not the last one)

.github/workflows/validate_urls.yml
Tyler-Keith-Thompson: That's really cool! The comment is still useful, since I know our maintainers mostly use VS code.

.github/workflows/validate_urls.yml
Richard-Gist: I was not taking a pre-emptive exclusion strategy.  This was the only URL that failed but passed when I manually navigated to the URL.

I'm thinking that we don't exclude until it fails.  If we pre-emptively exclude, we may inadvertently exclude things we would want to check.  For instance, if you just exclude issues, then a link to an important issue would be excluded from being checked.

Richard-Gist: All that to say, I don't think we should add any more excludes until they fail.

.github/workflows/validate_urls.yml
Richard-Gist: How 'bout `Branch to crawl for broken links in .md and .swift files`

Richard-Gist (commented):
Because I'm also a maintainer, I ended up starting a review myself.  Oops.

.github/workflows/validate_urls.yml
Tyler-Keith-Thompson: Honestly, despite the length, I appreciate the context. I like it better.

.github/workflows/validate_urls.yml
Tyler-Keith-Thompson: I can live with that.

.github/workflows/validate_urls.yml
Richard-Gist: I don't believe so in this job.  Specifically, because I don't believe there should be any .md files, and if there are, I don't know if they are shown to the viewer.  A quick scan of the `gh-pages` in my fork does not have any such files.

Tyler-Keith-Thompson: What about CSS files? Javascript? Anything necessary there?

Richard-Gist: Can you link from a CSS file?  Javascript I thought about, but I don't want to open that box of javascript linking, which would be generated.  I could, and we could see if it works but I have a feeling it would break depending on how the javascript does its navigations.

Tyler-Keith-Thompson: Yeah, you can with CSS, things like background images can be links, for example.

Tyler-Keith-Thompson: There's also this, which is more relevant than backgrounds: https://developer.mozilla.org/en-US/docs/Web/CSS/url()

Richard-Gist: In that case, we could add `.css`.

```suggestion
        file_types: .html,.css
```

Richard-Gist: ☝️ If that's all we need for .css scanning, commit.  But if I'm missing some feel free to make another suggestion.

Tyler-Keith-Thompson: It may be mildly overzealous but it's also cheap. I'd suggest
```suggestion
        file_types: .html,.css,.scss
```

Tyler-Keith-Thompson (approved):

PR #160, opened 2022-01-13
Preview Providers now compile correctly

### LINKED ISSUE #156: Allow `thenProceed` in preview providers
I want to use previews to view my `FlowRepresentable` as it'll appear in a workflow, even if that's just a single item workflow. However, right now `thenProceed` is not available.

Either make `thenProceed` globally available so we quit having one-off "I need it added" or add it to `PreviewProvider`.

I could write my own extension for this, but it should just work out of the box. Additionally, I suppose we could wrap `thenProceed` in some new protocol? However, I don't like the idea that I have to conform to a protocol just to define a workflow. There may be something to investigating resultBuilders in a future version of Swift where they allow those scoped functions. I've seen proposals about it, but for now I don't think we should go that route.

Tyler-Keith-Thompson:
## Checklist:

----

### DIFF
--- a/Sources/SwiftCurrent_SwiftUI/Extensions/ThenProceedExtensions.swift
+++ b/Sources/SwiftCurrent_SwiftUI/Extensions/ThenProceedExtensions.swift
@@ -24,6 +24,8 @@ private func verifyWorkflowIsWellFormed<LHS: FlowRepresentable, RHS: FlowReprese
  Adds an item to the workflow; enforces the `FlowRepresentable.WorkflowOutput` of the previous item matches the args that will be passed forward.
  - Parameter with: a `FlowRepresentable` type that should be presented.
  - Returns: a new `WorkflowItem` with the additional `FlowRepresentable` item.
+ - NOTE: Should be called inside a `WorkflowLauncher` initializer.
+ - IMPORTANT: Not for use in UIKit, unless you're doing SwiftUI interop.
  */
 @available(iOS 14.0, macOS 11, tvOS 14.0, watchOS 7.0, *)
 public func thenProceed<FR: FlowRepresentable & View>(with: FR.Type) -> WorkflowItem<FR, Never, FR> {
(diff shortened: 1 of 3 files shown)


morganzellers:
> * There is a breaking change in here. You used to be able to call `self.thenProceed`, which referred to a local method. Now that the method is global and the local methods are removed this no longer works and results in an error. @morganzellers what do you want to do about that?

I think we should version bump then. Does this force us to do a major version bump, or could we get away with bumping to v4.6.0?

Sources/SwiftCurrent_SwiftUI/Extensions/ThenProceedExtensions.swift
```diff
@@ -20,143 +19,56 @@ private func verifyWorkflowIsWellFormed<LHS: FlowRepresentable, RHS: FlowReprese
...
-        return WorkflowItem(VC.self) { nextItem() }
-    }
-    #endif
+/**
+ Adds an item to the workflow; enforces the `FlowRepresentable.WorkflowOutput` of the previous item matches the args that will be passed forward.
+ - Parameter with: a `FlowRepresentable` type that should be presented.
+ - Parameter nextItem: a closure returning the next item in the `Workflow`.
+ - Returns: a new `WorkflowItem` with the additional `FlowRepresentable` item.
+ - NOTE: Should be called inside a `WorkflowLauncher` initializer.
+ - IMPORTANT: Not for use in UIKit, unless you're doing SwiftUI interop.
```
morganzellers, line 67: Nice nice

morganzellers (approved):

morganzellers:
I feel like I'm okay just gating this for our next version bump. It feels far away, but this isn't worth bumping to 5.0 now

Tyler-Keith-Thompson:
Agreed. Should we create a 5.0 branch and merge it there?

morganzellers:
> Agreed. Should we create a 5.0 branch and merge it there?

I like that idea, I can do that

PR #169, opened 2022-02-03
Added open PR badge to exclusions.

morganzellers:
## Checklist:

### DIFF
--- a/.github/workflows/validate_urls.yml
+++ b/.github/workflows/validate_urls.yml
@@ -27,7 +27,7 @@ jobs:
         git_path: https://github.com/wwt/SwiftCurrent
         branch: ${{ env.GIT_BRANCH }}
         file_types: .md,.swift
-        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social
+        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social,https://img.shields.io/github/issues-pr/wwt/SwiftCurrent?color=bright-green
     - name: slack-send
       if: failure()
       uses: slackapi/slack-github-action@v1.16.0
@@ -46,7 +46,7 @@ jobs:
         git_path: https://github.com/wwt/SwiftCurrent
         branch: gh-pages
         file_types: .html,.css,.scss
-        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://github.com/realm/jazzy,https://realm.io,https://github.com/wwt/SwiftCurrent,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social
+        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://github.com/realm/jazzy,https://realm.io,https://github.com/wwt/SwiftCurrent,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social,https://img.shields.io/github/issues-pr/wwt/SwiftCurrent?color=bright-green
     - name: slack-send
       if: failure()
       uses: slackapi/slack-github-action@v1.16.0


Tyler-Keith-Thompson (commented):
So...we're adding a lot of shields here. I kind of get it because shields are badly rate limited but before I approve I feel compelled to ask the question are we doing the right thing by excluding all of these? Have we explored other options? Are we comfortable with them being down and not validated?

nickkaczmarek:
@Tyler-Keith-Thompson @morganzellers Looking into this a little deeper I'm not sure there is much more we can do to assure the uptime of the shields beyond hosting our own server for them (which is an option). 

> Have we explored other options?

If we wanted to go down the server hosting route it seems like we might be able to leverage a service like netlify or vercel to handle the hosting and might even be able to do so for free. I'm not sure if that would be more headache than it's worth.

Another option might be to just statically pull in the images and reference them in the README. I'm not sure what sort of issues this would cause, but it seems like we could effectively cache these since the shields are only displaying static info as opposed to CI status or code coverage.

> Are we comfortable with them being down and not validated?

I'm torn on this. I see them as a nice to have and while it would be good to know if they're unavailable, I'm not sure if having notifications that they're down without any ability to bring them back up is worth the headache. I'm kind of leaning towards the change @morganzellers made or downloading and caching them in our project.

I'm curious what y'all think? Maybe this is a longer discussion.

morganzellers:
@nickkaczmarek @Tyler-Keith-Thompson I think it may be worthwhile to download (upload?) the static ones to the repo and just exclude the dynamic ones from the checker. I think hosting our own thing is overkill for the badges, but at the same time I don't want to end up with all of our badges excluded and it feels like that's where this is headed.

nickkaczmarek:
@morganzellers do we need to exclude the dynamic badges? Like, are they as likely to go down?

nickkaczmarek:
> @nickkaczmarek @Tyler-Keith-Thompson I think it may be worthwhile to download (upload?) the static ones to the repo and just exclude the dynamic ones from the checker. I think hosting our own thing is overkill for the badges, but at the same time I don't want to end up with all of our badges excluded and it feels like that's where this is headed.

sideload? 😈

Tyler-Keith-Thompson:
The probably isn't really that shields.io goes down it's that dynamic badges poll the github APIs and they get rate limited. FWIW none of them are actually static? Some are just slow-moving. Like the platforms we support is dynamically pulled from our podspec.

nickkaczmarek:
@morganzellers Going to look into if we can regex add the shield images to exclude_urls, but if that doesn't pan out I will approve and merge this.

morganzellers:
> @morganzellers Going to look into if we can regex add the shield images to exclude_urls, but if that doesn't pan out I will approve and merge this.

Great idea, sounds good!

.github/workflows/validate_urls.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         git_path: https://github.com/wwt/SwiftCurrent
         branch: ${{ env.GIT_BRANCH }}
         file_types: .md,.swift
-        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social
+        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social,https://img.shields.io/github/issues-pr/wwt/SwiftCurrent?color=bright-green
```
nickkaczmarek, line 30:
```suggestion
        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new
        exclude_patterns: https://img.shields.io
```

.github/workflows/validate_urls.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         git_path: https://github.com/wwt/SwiftCurrent
         branch: gh-pages
         file_types: .html,.css,.scss
-        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://github.com/realm/jazzy,https://realm.io,https://github.com/wwt/SwiftCurrent,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social
+        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://github.com/realm/jazzy,https://realm.io,https://github.com/wwt/SwiftCurrent,https://img.shields.io/cocoapods/v/SwiftCurrent.svg?style=popout,https://img.shields.io/github/license/wwt/SwiftCurrent,https://img.shields.io/github/stars/wwt/SwiftCurrent?style=social,https://img.shields.io/github/issues-pr/wwt/SwiftCurrent?color=bright-green
```
nickkaczmarek, line 49:
```suggestion
        exclude_urls: https://github.com/wwt/SwiftCurrent/discussions/new,https://github.com/realm/jazzy,https://realm.io,https://github.com/wwt/SwiftCurrent
        exclude_patterns: https://img.shields.io
```

nickkaczmarek (changes requested):
Suggested the `exclude_patterns` style in order to be more concise.

morganzellers:
Great find @nickkaczmarek! This is great

nickkaczmarek (approved):

PR #202, opened 2022-04-11
Upping version of SwiftSyntax in the CLI to lastest.

morganzellers:
## Checklist:


Tyler-Keith-Thompson (changes requested):
This requires consumers to use Swift 5.6, why would we place that restriction on them?

morganzellers:
Testing these changes as local packages fixed the problems seen in the issue. Not sure what the difference is, but the version we were using was a handful of months behind. The [release page](https://github.com/apple/swift-syntax/releases/tag/0.50600.1) mentions vending the internal parser library as a binary framework on macOS, but 🤷🏻.

morganzellers:
> This requires consumers to have Swift 5.6, why would we place that restriction on them?

Ah, yeah I didn't consider that.

Tyler-Keith-Thompson (approved):
With the understanding that this change took us from "it doesn't work" to "it works" I think accepting Swift 5.6 makes sense.

PR #207, opened 2022-06-23
Add variation of XCTAssertWorkflowLaunched for use with AnyWorkflow

nallick:
## Checklist:

----

### If Applicable:

----

### DIFF
--- a/Sources/SwiftCurrent_Testing/CustomAssertions.swift
+++ b/Sources/SwiftCurrent_Testing/CustomAssertions.swift
@@ -23,11 +23,12 @@ public func XCTAssertWorkflowLaunched<F>(from VC: UIViewController, workflow: Wo
 
 /// Assert that a workflow was launched and matches the workflow passed in
 public func XCTAssertWorkflowLaunched(from VC: UIViewController, workflow: AnyWorkflow?, file: StaticString = #filePath, line: UInt = #line) {
-    let last = VC.launchedWorkflows.last
     guard let workflow = workflow else {
-        XCTAssertNil(last, "workflow found when none expected", file: file, line: line)
+        XCTAssertNoWorkflowLaunched(from: VC, file: file, line: line)
         return
     }
+
+    let last = VC.launchedWorkflows.last
     XCTAssertNotNil(last, "No workflow found", file: file, line: line)
     guard let listenerWorkflow = last,
           listenerWorkflow.count == workflow.count else {
@@ -44,6 +45,11 @@ public func XCTAssertWorkflowLaunched(from VC: UIViewController, workflow: AnyWo
         let expected = workflowNode.value.metadata.flowRepresentableTypeDescriptor
         XCTAssert(actual == expected, "Expected type: \(expected), but got: \(actual)", file: file, line: line)
     }
+}
+
+/// Assert that no workflow was launched
+public func XCTAssertNoWorkflowLaunched(from VC: UIViewController, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertNil(VC.launchedWorkflows.last, "workflow found when none expected", file: file, line: line)
 }
 #endif
 #endif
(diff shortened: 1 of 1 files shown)


Sources/SwiftCurrent_Testing/CustomAssertions.swift
```diff
@@ -18,7 +18,16 @@ import UIKit
...
 /// Assert that a workflow was launched and matches the workflow passed in
 public func XCTAssertWorkflowLaunched<F>(from VC: UIViewController, workflow: Workflow<F>, file: StaticString = #filePath, line: UInt = #line) {
+    XCTAssertWorkflowLaunched(from: VC, workflow: AnyWorkflow(workflow), file: file, line: line)
+}
+
+/// Assert that a workflow was launched and matches the workflow passed in
+public func XCTAssertWorkflowLaunched(from VC: UIViewController, workflow: AnyWorkflow?, file: StaticString = #filePath, line: UInt = #line) {
     let last = VC.launchedWorkflows.last
+    guard let workflow = workflow else {
+        XCTAssertNil(last, "workflow found when none expected", file: file, line: line)
```
Tyler-Keith-Thompson, line 28: If we're going to allow people to assert no workflow was launched I think it should be a separate method, otherwise this looks great!

nallick: Fair enough, although the workflow here is only optional as a convience for the case where data-driven workflow generation fails and comes back nil. If your test is generating a nil expected workflow, you'd hope no actual workflow launched.

Sources/SwiftCurrent_Testing/CustomAssertions.swift
```diff
@@ -18,6 +18,16 @@ import UIKit
 #if !os(watchOS)
 /// Assert that a workflow was launched and matches the workflow passed in
 public func XCTAssertWorkflowLaunched<F>(from VC: UIViewController, workflow: Workflow<F>, file: StaticString = #filePath, line: UInt = #line) {
```
Tyler-Keith-Thompson, line 20: Perhaps this workflow should be optional, given the other version is?

Tyler-Keith-Thompson:
@nallick This looks good. I'll need you to sign the CLA before I can approve it.

Tyler-Keith-Thompson (approved):

Sources/SwiftCurrent_Testing/CustomAssertions.swift
nallick: I don't think that would work here since it's generic on the parameter value. If you passed in nil it wouldn't know what the generic type should be. Plus it wouldn't be able to tell which overloaded function to call, since for nil the two signatures would be identical.

nallick: I suppose another way to look at it would be to consider that this variant would be used for workflows created programatically. Do you envision a case where a programatic workflow might be optional?

nallick:
> @nallick This looks good. I'll need you to sign the CLA before I can ~approve~ merge it.

There's something weird going on with the CLA tool. I signed it at the same time as I created the PR, but it doesn't show here. When I try to sign it again I get the page with ````You have agreed to the CLA for wwt/SwiftCurrent```` and the signing controls at the bottom disabled.

nallick:
Screen Shot 2022-06-27 at 11 50 26 AM (2)

Tyler-Keith-Thompson:
> > @nallick This looks good. I'll need you to sign the CLA before I can ~approve~ merge it.
> 
> @Tyler-Keith-Thompson There's something weird going on with the CLA tool. I signed it at the same time as I created the PR, but it doesn't show here. When I try to sign it again I get the page with `You have agreed to the CLA for wwt/SwiftCurrent` and the signing controls at the bottom disabled. Also clicking the "recheck" button doesn't make a difference.
> 
> Screen Shot 2022-06-27 at 11 50 26 AM (2)

I think the problem is you contributed with 2 accounts. One is "nallick" and the other is "nick-nallick-wwt". Both need to sign the agreement for it to work.

Sources/SwiftCurrent_Testing/CustomAssertions.swift
Tyler-Keith-Thompson: The only case I can dream up is if somebody either does `nil as Workflow<Type>` or has `let wf: Workflow<Type>?` and assigns it later. The compiler should disambiguate `nil` to the most reasonable candidate, which is `AnyWorkflow`.

PR #211, opened 2022-08-13
Add the ability to abandon workflow when a publisher emits a value

### LINKED ISSUE #95: SwiftUI: Abandon from outside `FlowRepresentable`
I've got multiple workflows in SwiftUI being presented at the same time. I'd like to abandon them all if a user is logged out. This is achievable in UIKit because you can just extract and use the data model. For example, I could've created multiple workflows and stored them somewhere, then just called abandon on them. However, because SwiftUI does not use the data model, that trick doesn't work.

Realistically this is yet another example in my head of why we should allow the data model to be used in SwiftUI however this feature request is different. I'd like some way of abandoning a workflow from outside a `FlowRepresentable` think something like this:
```swift
WorkflowLauncher(isLaunched: .constant(true))
                .thenProceed(with: WorkflowItem(LoginView.self))
                .thenProceed(with: WorkflowItem(QRScannerFeatureOnboardingView.self))
                .thenProceed(with: WorkflowItem(QRScannerFeatureView.self))
                .abandonOn(user.isLoggedIn == false) // Version 1, Autoclosure
                .abandonOn { !user.isLoggedIn } // Version 2, for reals closure
                .tabItem {
                    Label("QR Scanner", systemImage: "camera")
                }
                .tag(Tab.qr)
```
That exact solution may not actually be possible, but you can see the advantage of describing this here.

#### Allow me to use data models:
This solution is a huge request, but it carries some very reasonable possibilities for me outside this one use-case, for example:
```swift
@State var workflow = Workflow(LoginView.self)
                                        .thenProceed(with: QRScannerFeatureOnboardingView.self)
                                        .thenProceed(with: QRScannerFeatureView.self)
                                        
// In the view
WorkflowLauncher(isLaunched: .constant(true))
                .launchInto(workflow)
                .tabItem {
                    Label("QR Scanner", systemImage: "camera")
                }
                .tag(Tab.qr)
                .onChange(of: user.isLoggedIn) { if !$0 { workflow.abandon() } } // abandons and relaunches, yay!
```

#### With how things are now:
The current alternative is for me to go to literally every `FlowRepresentable` view and add an `onChange` modifier that abandons the workflow if the user is not logged in anymore. This is highly unfortunate because then every view has to know about the user, even if it would otherwise not need to.

### Additional context
There may be completely new options or patterns I have not yet thought through. I'm happy to have a discussion about that if my premise is incorrect.

Tyler-Keith-Thompson:
## Checklist:

### DIFF
--- a/Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
+++ b/Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
@@ -8,6 +8,7 @@
 
 import XCTest
 import SwiftUI
+import Combine
 import ViewInspector
 
 import SwiftCurrent
@@ -527,6 +528,32 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
         XCTAssertThrowsError(try launcher.find(FR1.self))
 
         wait(for: [expectOnAbandon1, expectOnAbandon2], timeout: TestConstant.timeout)
+    }
+
+    func testWorkflowCanAbandonFromAPublisher() async throws {
+        let publisher = PassthroughSubject<Void, Never>()
+        struct FR1: View, FlowRepresentable, Inspectable {
+            var _workflowPointer: AnyFlowRepresentable?
+            var body: some View { Text("FR1 type") }
+        }
+        let isLaunched = Binding(wrappedValue: true)
+        let expectOnAbandon = expectation(description: "OnAbandon called")
+        let launcher = try await MainActor.run {
+            WorkflowView(isLaunched: isLaunched) {
+                WorkflowItem(FR1.self)
+            }
+            .onAbandon {
+                XCTAssertFalse(isLaunched.wrappedValue)
+                expectOnAbandon.fulfill()
+            }
+            .abandonOn(publisher)
+        }.hostAndInspect(with: \.inspection)
+
+        XCTAssertEqual(try launcher.find(FR1.self).text().string(), "FR1 type")
+        publisher.send()
+        XCTAssertThrowsError(try launcher.find(FR1.self))
+
+        wait(for: [expectOnAbandon], timeout: TestConstant.timeout)
     }
 
     func testWorkflowCanHaveModifiers() async throws {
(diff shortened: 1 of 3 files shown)


Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
```diff
@@ -529,6 +530,32 @@ final class SwiftCurrent_SwiftUIConsumerTests: XCTestCase {
         wait(for: [expectOnAbandon1, expectOnAbandon2], timeout: TestConstant.timeout)
     }
 
+    func testWorkflowCanAbandonFromAPublisher() async throws {
+        let publisher = PassthroughSubject<Void, Never>()
```
Richard-Gist, line 534: So other ways of using this would be to have an error publisher that's Output is Error and Failure is Never.  Then you could tie your workflow to automatically abandon if the publisher publishes an error.

I think that means that in general you subscribe to the publisher you really care about in your code, process the results and then send from the publisher you gave to the workflow.  Something like: RealPublisher -> ProcessPublisher -> Workflow publisher.

I'm trying to think through how this would effectively be used in the wild and how this could address the issue you are closing.

Richard-Gist (commented):
I think this is my only outstanding issue right now.  If I can get my head wrapped around it, I'm good to approve.

Tests/SwiftCurrent_SwiftUITests/SwiftCurrent_SwiftUITests.swift
Tyler-Keith-Thompson: So I'll state it more like this:
> The publisher you can pass to WorkflowView cannot have a failure type, because there's no logical behavior for dealing with errors.

What this means:
If you use a publisher (the "Real Publisher") that doesn't fail, you're already good. An example might be a publisher you have tied into button presses or observed objects with `@Published` properties. This covers an already large number of publishers you might be using. 

What's cool is that the `WorkflowView` will abandon when any value is emitted, so you don't have to do much other than hook it up to something that emits values.

What if your publisher can fail?
Then *you* must add explicit logic. One way to do that is simply to transform the error. For example I could take the "Real" publisher and call `.replaceError(with: <SomeValue>)`. 

Alternatively, maybe what you want is to trigger off of state. for that you could use `onChange`. In my example of abandoning a workflow when a user is logged out you could do something like.

```swift
VStack {
    Text("My View")
}
.onChange(User.isLoggedIn) { 
    if !$0 { 
        abandonPublisher.send() 
    } 
}
```

For this you need the setup you were just describing, a publisher you create yourself.

Does that help?

Richard-Gist: Yup thank you!

